Issue #9, opened 2017-02-13
macOS: error encountered while building uuid


automactic:
Here is the log

[uuid_error.txt](https://github.com/kiwix/kiwix-build/files/771608/uuid_error.txt)

mgautierfr:
This is a bug in e2fsprogs (uuid).
The `quota_inode_truncate` use a simple `return` to quit the function but it should return `errcode_t`.
On linux, gcc print a warning but the default compiler on macOS (clang ? gcc ?) see it as an error (which it is).

We use a pretty old version of e2fsprogs (1.42). This bug is fixed in recent version (since 1.42.5). The last version of e2fsprogs at this time is 1.43.4.

We should move to the last version.

automactic:
The default compiler on macOS now is clang

mgautierfr:
This last commit should fix this problem.
You may want to remove the old `uuid-1.42` in `ARCHIVE`, `SOURCE` and `BUILD_native_static` but is not necessary. Just relaunch the `kiwix-build.py` with same arguments.

automactic:
I just re-runed the same command, but I got SSL cert error while re-downloading uuid

automactic:
I removed everything related to `uuid-1.42`. I still got ssl error. Here's the log:
[uuid_ssl_error.txt](https://github.com/kiwix/kiwix-build/files/771764/uuid_ssl_error.txt)

mgautierfr:
Hum, that's strange, I don't have this error.
Are you behind a proxy ? Or something that could "man in the middle" your connection to `https://www.kernel.org` ?

(I've just push a new version about the uuid, I had forget to change the uuid version. But it should not be relevant to your ssl error)

automactic:
I have no idea why. Even doing
```
import urllib.request
response = urllib.request.urlopen('https://www.google.com')
```
results in the same ssl error. I will try again later when I got home.

automactic:
I cannot solve it successfully. Based on a quick google search, I think I am not alone in these kind of problem (macOS + urllib + cert issue)

I am able to use the following code to bypass cert verify
```
import ssl
import urllib 
context = ssl.create_default_context()
context.check_hostname = False
context.verify_mode = ssl.CERT_NONE
response = urllib.request.urlopen('https://www.google.com', context=context)
```

kelson42:
I think we need here a --no-cert-check option to the script.

automactic:
[...]

PR #13, opened 2017-03-01
Support flag --no_cert_check

automactic:
add support for `--no_cert_check` flag, trying to solve #9

### DIFF
--- a/kiwix-build.py
+++ b/kiwix-build.py
@@ -2,6 +2,7 @@
 
 import os, sys, stat
 import argparse
+import ssl
 import urllib.request
 import tarfile
 import subprocess
@@ -399,7 +400,14 @@ class BuildEnv:
             if what.sha256 == get_sha256(file_path):
                 raise SkipCommand()
             os.remove(file_path)
-        urllib.request.urlretrieve(file_url, file_path)
+        if options.no_cert_check == True:
+            context = ssl.create_default_context()
+            context.check_hostname = False
+            context.verify_mode = ssl.CERT_NONE
+            with urllib.request.urlopen(file_url, context=context) as resource, open(file_path, 'wb') as file:
+                file.write(resource.read())
+        else:
+            urllib.request.urlretrieve(file_url, file_path)
         if not what.sha256:
             print('Sha256 for {} not set, do no verify download'.format(what.name))
         elif what.sha256 != get_sha256(file_path):
@@ -1023,6 +1031,8 @@ def parse_args():
     parser.add_argument('--verbose', '-v', action="store_true",
                         help=("Print all logs on stdout instead of in specific"
                               " log files per commands"))
+    parser.add_argument('--no-cert-check', action='store_true',
+                        help="Skip SSL certificate verification during download")
     return parser.parse_args()
 
 if __name__ == "__main__":
--- a/kiwix-build.py
+++ b/kiwix-build.py
@@ -400,14 +400,21 @@ class BuildEnv:
             if what.sha256 == get_sha256(file_path):
                 raise SkipCommand()
             os.remove(file_path)
+
         if options.no_cert_check == True:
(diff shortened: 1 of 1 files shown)


kiwix-build.py
```diff
@@ -399,7 +400,14 @@ def download(self, what, where=None):
...
             os.remove(file_path)
-        urllib.request.urlretrieve(file_url, file_path)
+        if options.no_cert_check == True:
+            context = ssl.create_default_context()
+            context.check_hostname = False
+            context.verify_mode = ssl.CERT_NONE
+            with urllib.request.urlopen(file_url, context=context) as resource, open(file_path, 'wb') as file:
+                file.write(resource.read())
+        else:
+            urllib.request.urlretrieve(file_url, file_path)
```
mgautierfr: I would prefer to have only "way" (line) to effectively download the file (and so reduce different code to do the same thing)

Something like that would be better:
```python
if options.no_cert_check == True:
    # Set option to not check certificate
    context = ...
else:
    # Set option to check certificate
    context = ...
with urllib.request.urlopen(file_url, context=context) as resource, open(file_path, 'wb') as file:
    # Write resource data into file.
    ...
```

kiwix-build.py
```diff
@@ -399,7 +400,14 @@ def download(self, what, where=None):
             if what.sha256 == get_sha256(file_path):
                 raise SkipCommand()
             os.remove(file_path)
-        urllib.request.urlretrieve(file_url, file_path)
+        if options.no_cert_check == True:
+            context = ssl.create_default_context()
+            context.check_hostname = False
+            context.verify_mode = ssl.CERT_NONE
+            with urllib.request.urlopen(file_url, context=context) as resource, open(file_path, 'wb') as file:
+                file.write(resource.read())
```
mgautierfr: Here we will read all the content at once and write it once in the file.
On big archive, this meanr that we will use a lot of memory just to handle a temporary buffer.

It should be better to copy the content by small chunks, ones after the others.

For most of dependencies, archives are pretty small and it is not a problem but android tools (ndk, sdk) are pretty big (more than 600Mo)

(To be honest, I don't know if `urlretrieve` handles this problem correctly, I cannot find any clues in the documentation. But now, if we handle inner mechanics ourselves, we have to handle this point)

automactic: I think `urlretrieve` is more likely to be the one that does not store the file in memory and flush to disk, but it does not support opting out of ssl check. Why don't we keep both for now?

mgautierfr: > Why don't we keep both for now?

Because I don't want to comeback on this point latter :) The less code branches there is, the better it is.

This should not be to difficult to implement, a small read in a loop should be enough.
I've check `urlretrieve` and this is globally what it does (https://github.com/python/cpython/blob/master/Lib/urllib/request.py#L229). It reads small chunk of 8Ko in a `while` loop.

automactic: Are you saying we should create our own batched download function that also support bypass SSL verification?

mgautierfr: Well maybe not create a **new** function just for this.
But the function we are changing is already a `download` function. We just have to add a loop around the `read` to make it works with small chunks.

mgautierfr:
Your changes seem good.
Can you just squash your two commits into one and rebase your PR on the `origin/master` ?
Then I will merge it.

automactic:
Did I do the correct thing? Can I delete `feature--no-cert-check` now?

mgautierfr:
I would have prefer you to rebase and `push --force` `feature--no-cert-check` to allow me to have a last look on it and merge it myself*.
But anyway, what you've merged is correct, so no need to revert it. Thanks for your patch.

*It's not I especially want to be the one how merge PR, but I prefer that the one merging a PR is not the same who have written it. And it is the same in the opposite directory, do not hesitate to review my PRs.

PR #14, opened 2017-03-02
Android

mgautierfr:
Add support for android cross-compilation.

It also add a lot of change in the way we handle cross-compilation.
New cross-compilation (to raspeberryPi and to iOS from macOS) should be easier to implement.


mgautierfr:
This branch seems to work well. It also add a lots of chane in the kiwix-build.py structure.
It's better to merge it right now and base other changes on top of it.

PR #16, opened 2017-03-07
Correct option name "--build-static" in readme

dattaz:


mgautierfr:
Thanks for your patch but I've recently merged the `android` branch into master and it changes the options of kiwix-build.
It add new build platforms and the notion of static build was a non sense for some of them.
So there is no more `--build-static` option.

PR #21, opened 2017-03-21
Zimwriterfs

mgautierfr:
Add support for zimwriterfs in kiwix-build.py

### DIFF
--- a/dependencies.py
+++ b/dependencies.py
@@ -224,14 +224,15 @@ class Icu_cross_compile(Icu):
             return super().configure_option + " --with-cross-build=" + Icu_native.builder.build_path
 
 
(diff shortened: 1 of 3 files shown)


kelson42:
@mgautierfr I do no really understand why kiwix has a build dependency to zimwriterfs?

mgautierfr:
`Dependency` is not the best name. But it is the name I used in kiwix-build.py.
It should be called `AnyProjectKiwixBuildCanBuild` but it's not a better name :)

This PR mainly add the possibility to build the zimwritterfs with kiwix-build.py.

BTW, Why have you merged the `master` branch into this `zimwriterfs` branch ?

PR #23, opened 2017-04-06
Added the Debian automake package to the dependencies.

julianharty:
Please check and approve this change to add the automake and related packages so they're added as debian packages.


mgautierfr:
Thanks for the patch.

Issue #6, opened 2017-01-25
Add Linux/ARM to CI build targets


tim-moody:
Please include support for arm, especially rpi in builds

mgautierfr:
Ho yes, this is absolutely needed, and more than the cross-compilation for Windows we already have. 
It should be not complicated to add. I will do it in the week.

kelson42:
https://github.com/kiwix/kiwix-tools/issues/13 depends on that ticket

mgautierfr:
Closed with #30

kelson42:
Here is the first ARM nightly build http://download.kiwix.org/nightly/2017-04-29/kiwix-tools_armhf_2017-04-29.tar.gz

PR #30, opened 2017-04-14
Arm

mgautierfr:


.travis.yml
```diff
@@ -19,6 +19,8 @@ env:
  - PLATFORM="native_static"
  - PLATFORM="win32_dyn"
  - PLATFORM="win32_static"
+ - PLATFORM="raspberrypi_dyn"
+ - PLATFORM="raspberrypi_static"
```
kelson42: Would not be a better idea to talk about arm instead of raspberrypi? We have many other devices using arm with linux.

dependency_utils.py
```diff
@@ -134,7 +134,7 @@ def _git_clone(self, context):
         context.force_native_build = True
         if os.path.exists(self.git_path):
             raise SkipCommand()
-        command = "git clone " + self.git_remote
+        command = "git clone {} {}".format(self.git_remote, self.git_dir)
```
kelson42: Might make sense to use option --depth=1 to git clone command.

kelson42:
Really nice to see that merge request, here are the tickets which can be closed at the time it works:
* https://github.com/kiwix/kiwix-build/issues/6
* https://github.com/kiwix/kiwix-tools/issues/13

.travis.yml
mgautierfr: Hum, I've called it raspberrypi because I've used the compiler provided by raspberry team (gcc-linaro-arm-linux-gnueabihf-raspbian)

This is a compiler based on raspbian but I think we can assume that generated binaries will run on other distributions / armhf platforms.

I will rename the platform `armhf` instead of `raspberrypi`.

dependency_utils.py
mgautierfr: I was about to tell you it was necessary to get the full history to be able to commit and push updates but it is no more the case \o/

Recent version of git 1.9 (available in trusty) allow to commit and push from shallow repository.
And more recent version (2.8) even allow to unshallow a repository (it is not available in trusty... but I don't care :) )

I will change this, but not in this PR, this is not relevant to the arm functionality (the commit itself is not relevant to arm, so I will remove it from the PR and create another one)

mgautierfr:
I've update the code accordingly to your remarks.
Thanks for your review.

tim-moody:
Great. I'll give it a try when I'm back home.

Issue #3, opened 2017-01-18
Improve kiwix-build README


kelson42:
README should clearly explain the compilation procedure. This means step-by-step procedures which can be follow blindly with short explanation about what is going on:
* What are the dependencies (with minimal version)
* How to compile kiwix-lib and kiwix-tools under Linux static/dynamic by downloading locally deps
* How to cross-compile kiwix-lib and kiwix-tools under Linux for Windows static/dynamic by downloading locally deps

mgautierfr:
I've merged the branch `android` into `master` and with that, I've rewrite the README. Can you check it on your side (and push changes in a new PR if necessary) ?

kelson42:
@mgautierfr thx, will have a look.

kelson42:
@mgautierfr I have started to rewrite the README and I have a few questions:
* How to get the list of target? How to compile a Kiwix APK for Android?
* How to compile for linux arm?

ynikitenko:
Please write that `qmake` must be installed. I had an error in the middle of building that qmake couldn't be found. I think that could be better reported in the very beginning (or, in a worse case, in README).
Also ninja could be automatically installed from pip. Why don't we add that to setup.py requirements? I think that should work fine in most cases.

ynikitenko:
This was suggested earlier, but the issue was closed (https://github.com/kiwix/kiwix-build/issues/49). I think there are two options: either do that automatically, or include in README (as proposed in this issue).

kelson42:
> Please write that `qmake` must be installed. I had an error in the middle of building that qmake couldn't be found. I think that could be better reported in the very beginning (or, in a worse case, in README).
> Also ninja could be automatically installed from pip. Why don't we add that to setup.py requirements? I think that should work fine in most cases.

@ynikitenko `qmake` is de dependence only used for Kiwix Desktop. Qt is a clear dependence (written in README.md) of Kiwix Desktop already. We don't repeat here what is written in specific project READMEs.

mgautierfr:
> So it seems that kiwix-build promises to deal with all dependencies itself? It would be great.

Yes. And that is THE purpose of kiwix-build. But we are speaking about run-time dependency (either by compiling them or using the system packages)
For build dependencies (qmake, cmake, ninja,..) we use the system packages.

There is already a system to install needed packages (deb, rpm). If some needed packages are not installed by kiwix-build, please report it in new issues.

ynikitenko:
[...]

PR #39, opened 2017-04-23
Update README on how to build + auto-install pkg-config package 

mossroy:
+ fix a few typos

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,12 @@ of kiwix.
 Most of the compilation steps are handled by the kiwix-build.py
 script.
 
-This script has been tested of Fedora 23 and Ubuntu 16.10
+This script has been tested on Fedora 23 and Ubuntu 16.10
 
 If you want to cross-compile for other platforms as windows or android, you
 should read the "Other target platform" section.
+
+This script is not officially supported on Debian 8, but has been reported to work, after adding ~/.local/bin to the PATH.
 
 ## Preparing environment
 
@@ -28,14 +30,16 @@ them with your package manager.
 Else you can install them manually
 
 ### Meson
+You need to install python 3 and pip. On debian and ubuntu, they are provided by the package python3-pip
 
 ```
-pip install meson --user # Install Meson
+pip3 install meson --user # Install Meson
 ```
 
 (You may want to install meson in a virtualenv if you prefer)
 
 ### ninja
+You need git to be installed
 
 ```
 git clone git://github.com/ninja-build/ninja.git
@@ -54,10 +58,10 @@ This is the simple one.
 ```
 
 It will compile everything.
-If you are using a supported platform (redhad or debian based) it will install
+If you are using a supported platform (redhat or debian based) it will install
 missing packages using sudo.
 If you don't want to trust kiwix-build.py and give it the root right, just
-launch yourself the the printed command.
+launch yourself the printed command.
 
 ### Outputs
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ of kiwix.
 Most of the compilation steps are handled by the kiwix-build.py
 script.
 
-This script has been tested of Fedora 23 and Ubuntu 16.10
+This script has been tested on Fedora 23, Ubuntu 16.10 and Debian 8
```
kelson42: Who assures that this will be always tested on Debian?

kelson42:
Thx for the fixes. Kind of waiting that the Android compilation problems are finally fixed with Kiwix build to revamp that README anyway. See https://github.com/kiwix/kiwix-build/issues/3.

README.md
mossroy: Well, I tested on Debian 8 with current version, but will certainly not test on every release.

mgautierfr: In this case we should not say it is tested on Debian 8.
This sentence is a kind of implicit contract. It implies that the script is intended to work on Fedora 23 and Ubuntu 16.10 (and that we will fix potential bug)

README.md
```diff
@@ -28,14 +28,17 @@ them with your package manager.
 Else you can install them manually
 
 ### Meson
+You need to install python 3 and pip. On debian and ubuntu, they are provided by the package python3-pip
 
 ```
-pip install meson --user # Install Meson
+pip3 install meson --user # Install Meson
+cp ~/.local/bin/meson <a_directory_in_your_path>
```
mgautierfr: Does `~/.local/bin` not already in the `$PATH` ?
It's seems to be the case on Fedora and Ubuntu. (And that's way `pip` install binaries in this directory)

README.md
```diff
@@ -48,16 +51,17 @@ cp ninja <a_directory_in_your_path>
 ## Buildings
 
 This is the simple one.
+You need pkg-config to be installed
```
mgautierfr: It would be better to make kiwix-build install the pkg-config package if needed.
Adding the package name in the `COMMON` entries of the `PACKAGE_NAME_MAPPERS` dictionaries in `kiwix-build.py` will be enough.

README.md
mossroy: Apparently, it has been added for next version of Debian (Stretch) : https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=820856
For Ubuntu, it has been added in 16.10, and backported on 16.04 : https://bugs.launchpad.net/ubuntu/+source/bash/+bug/1588562
So, If we do not want to officially support Debian 8 or earlier versions of Ubuntu, it's not worth mentioning, I guess.

README.md
mossroy: All right. I suppose I should add it to debian_native_dyn and debian_native_static, not for the other environments?

README.md
mossroy: I understand.
Maybe, we could add another sentence to say that Debian 8 is not officially supported, but has been reported to work after adding ~/.local/bin in the PATH ?

mossroy:
I rebased the commit and tried to take your comments into account

kelson42:
This is superseeded by https://github.com/kiwix/kiwix-build/pull/60. I close therefore that one.

@mossroy I have taken parts of your fixes and an other part (the pkg-config) had already been merged to master.

PR #40, opened 2017-04-24
Update ctpp2 patches to handle mingw32 compilation with meson 0.40.0

mgautierfr:
It seems that last meson version (0.40.0) do not pass cross compilation
extra_args to compiler when checking for headers.
Thus, the kiwix-lib check for ctpp2 headers fails because the WIN32
define is not present. So kiwix-lib is compiled without CTPP2 but
kiwix-tool compile as if kiwix-lib where compile with CTPP2 and fails.

By making CTPP2 headers check for `__WINGW32__` we workaround this problem.

I don't know if this is a bug in meson, ctpp2 or kiwix-lib but this small
fix allow to compile correctly with last meson version.

### DIFF
--- a/patches/ctpp2_mingw32.patch
+++ b/patches/ctpp2_mingw32.patch
@@ -83,3 +83,27 @@ diff -ur ctpp2-2.8.3/src/functions/FnRandom.cpp ctpp2-2.8.3.patched/src/function
  	srandomdev();
  #else
  	srandom(time(NULL));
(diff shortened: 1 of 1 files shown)


kelson42:
OK, @mgautierfr can you please open a ticket on Meson side, this looks like a regression!?

mgautierfr:
Meson issue open here : https://github.com/mesonbuild/meson/issues/1665

nirbheek:
This will be fixed in meson 0.40.1, thanks for reporting it!

PR #44, opened 2017-05-03
Icu custom data

mgautierfr:


mgautierfr:
Please @kelson42 review this PR. Especially this commit https://github.com/kiwix/kiwix-build/pull/44/commits/1109baec07f942548db576f684748aa961c78a5f

It's was you who has generated the previous `icudata56l.dat` so you have a clue on what to keep and what can be dropped.

dependencies.py
```diff
@@ -184,7 +184,8 @@ class Source(SvnClone):
         svn_dir = "icu4c"
 
         patches = ["icu4c_fix_static_lib_name_mingw.patch",
-                   "icu4c_android_elf64_st_info.patch"]
+                   "icu4c_android_elf64_st_info.patch",
+                   "icu4c_custom_data.patch"]
```
kelson42: What is the problem here, why do we need to patch ICU?

kelson42:
@mgautierfr In general looks good.

kelson42:
The only thing I can say about it comes from this comment:
https://github.com/kiwix/kiwix/blob/master/src/dependencies/Makefile.am#L247

```
# This files gathers ICU related data. It's a custom one generated
# with http://apps.icu-project.org/datacustom/ICUData56.html. It
# contains Break iterators, Collators and Transliterators.
```

PR #61, opened 2017-07-03
Git update now updates to latest commit

mhutti1:
When testing I discovered that kiwix-android is not updated on each run.

### DIFF
--- a/dependency_utils.py
+++ b/dependency_utils.py
@@ -145,6 +145,7 @@ class GitClone(Source):
             self.git_ref)
         self.buildEnv.run_command(command, self.git_path, context)
         self.buildEnv.run_command("git checkout "+self.git_ref, self.git_path, context)
+        self.buildEnv.run_command("git pull", self.git_path, context)
 
     def prepare(self):
         self.command('gitclone', self._git_clone)


mgautierfr:
> When testing I discovered that kiwix-android is not updated on each run.

Ho, you're right. That is bad and must be fixed.

However :
- In a general way, I really want that kiwix-build do not overwrite any changes made by a user in the source directory. This way a user (me 😄) can patch a source and run kiwix-build to test without lost its work.
- I dislike a lot `git pull`. The default behavior of `git pull` is to fetch the last version from remote (and we are already doing `git fetch`) and **merge** the origin branch into the local one. Most of the time, if you have local change commited (but not pushed), you will want to do a `rebase` and not a `merge`.

I would prefer something that check that no modification is made on the source tree (working tree clean, and no "new" branch currently checkout) and SKIP if it the case. And if every thing is right, launch a `git merge --ff-only`

kelson42:
@mhutti1 Anything blocking for you to go ahead on that branch?

mhutti1:
Haven't got round to it yet.

mhutti1:
@mgautierfr How about using https://stackoverflow.com/questions/3216360/merge-update-and-pull-git-branches-without-using-checkouts with the --update-head-ok flag?

kelson42:
Could we go ahead on that PR?

mgautierfr:
Superseded by #413

PR #62, opened 2017-07-05
Adding a few changes for Apple platforms

automactic:


mgautierfr:
The PR you're proposing is in fact twice the same (it seems) branch merged together.
Please "unmerge" it.

mgautierfr:
Even if I look in one single branch, there is a lot of commit fixing the precedent ones.
Those changes are useless (They may be useful while you are coding, but totally useless to keep).

For example:
 - e4d6f89 add a `Darwin_native_dyn` entry, e4d6f89 add a second identical and 8f1a451 remove the duplicate (while the commit comment says something else)
 - 0dd28f5 and dcd7071 do the exact contrary, one after the other. They can be simply dropped.

On the code itself :
- You've added a `Darwin_native_dyn` but no `Darwin_native_static` ? Is it because there is no "static" compilation on Darwin ?
- I suppose that only `native_dyn` and `iOS_arm64` is working on `Darwin` host. It would be good to add a check at `buildEnv` creation that the target_platform is working on the current host.

automactic:
Forgive my innocence, I have no idea why there could be identical commits and was quite confused when trying to remove them.

Can I simplify just consolidate all changes in `apple-platforms` into one commit and submit a pull request?

Regarding why there is no `Darwin_native_static`, that's because there is no need to build statically.

mgautierfr:
> Forgive my innocence, I have no idea why there could be identical commits.

No problem, this kind of "mistake" arrive more often that we think :)

> Can I simplify just consolidate all changes in apple-platforms into one commit and submit a pull request?

Yes, the changes are pretty small finaly. There is no need to keep several commits.

> Regarding why there is no Darwin_native_static, that's because there is no need to build statically.

Ok. If you feel up to add a check to avoid to build "invalid" targets (native_static, android_*, ...) on Darwin, please provide the changes in this PR (but separated commit). Else I will do it myself, no problem.

automactic:
@mgautierfr, I have squashed commits, please check. 

I am not sure where should I add checks in `BuildEnv`. Could you please add them? Thanks

kelson42:
@automactic @mgautierfr What happens here? Who needs to fix the the CI problem?

mgautierfr:
@automactic, I've just merged the branch. I was waiting that other branches about `ctpp2` were merged before.

> I am not sure where should I add checks in BuildEnv. Could you please add them? Thanks

Done

kelson42:
@mgautierfr thx

Issue #59, opened 2017-06-27
Impossible to build custom apps


kelson42:
It seems there is a bug with Travis and the Requests. I have open a ticket here:
https://github.com/travis-ci/travis-ci/issues/7982

See also #58 for the improvements.

kelson42:
@mgautierfr Upstream bug as been close as they think the problem as been resolved. Could you please confirm it still occurs and re-open it?

kelson42:
Custom app CI+publication will be handle within 3.0 milestone

kelson42:
We don't build Kiwix Android (custom apps) with Kiwix build.

PR #65, opened 2017-07-17
[CUSTOM_APP] Add a zim_path option to build_custom_app.py

mgautierfr:
If the user what to make a custom_app locally, he will need to download
the zim file before. As he has already the zim, it is better to give
the zim_path instead of give the zim_url and have `build_custom_app.py`
download a second time the zim.

### DIFF
--- a/build_custom_app.py
+++ b/build_custom_app.py
@@ -53,6 +53,7 @@ def parse_args():
     advance.add_argument('--extra-code', type=int, default=0)
     advance.add_argument('--check-certificate', default=True)
     advance.add_argument('--zim-url', default=None)
+    advance.add_argument('--zim-path', default=None)
     advance.add_argument('--no-android-upload', action='store_false', dest='android_upload')
 
     # Hidden options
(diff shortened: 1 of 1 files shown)


kelson42:
All this is extremely confusing.

kelson42:
Here is how it should work:
1 - custom app are build from json files, so customize your file if you want
2 - if you really want to give parameter on the command line, a generic option should be provided to overwritte any json field
3 - zim_path/zim_url is the same, remove one of them (like it was before) and download if this is a http link (with continue option)

mgautierfr:
The script is intended to be run on travis. All options being automatically set by using the json file in the kiwix/kiwix-android-custom repository.

This option is needed as the build using travis is broken because of #59 and we use this option only as workaround.

If this seems too confusing, I can hide the option itself. (this is already the case for a lot of options used in the command described [here](https://github.com/kiwix/kiwix-build/wiki/Build-and-publish-a-custom-app-locally-(without-travis)#upload-apks) )

> zim_path/zim_url is the same, remove one of them (like it was before) and download if this is a http link (with continue option). Otherwise it's really unclear which has the priority etc...

A path is not the same thing that a url. The information in zim_url in the json file will be used by script running in travis. Having a local path will not work.

mgautierfr:
I'm closing this PR as #66 supersedes this one and it should not be merged.

However, fell free to continue the discussion here or on #66.

PR #74, opened 2017-08-15
Add missing xapian patch.

automactic:
This is needed to have xapian generate a correct pkgconfig file if
linking with uuid is not necessary.


kelson42:
@automactic @mgautierfr Can we close that PR, it seems to me this is totally outdated?

mgautierfr:
Yes. It is outdated. I'm closing the PR

PR #86, opened 2017-10-17
Add support of `zim-tools` in kiwix-build.

mgautierfr:


kelson42:
@mgautierfr I do not know if you are OK with that branch, but if yes, please go ahead.

PR #109, opened 2018-01-24
Update to last version of icu4c.

mgautierfr:
With last version of glibc 2.26, `xlocale.h` is not provided anymore
(https://sourceware.org/glibc/wiki/Release/2.26#Removal_of_.27xlocale.h.27)

Icu has been patched (http://bugs.icu-project.org/trac/ticket/13385) so
we need to use the last version of icu.


mgautierfr:
This PR is replaced by #111

PR #140, opened 2018-04-01
Increment gradle wrapper version

mhutti1:


mhutti1:
This seems to fix the build but we get https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received. I think we might be because the build had to recompile dependencies but we should probably look into using the travis wait command.

mhutti1:
Restarting some of the builds helped but some still timeout

dependency_versions.py
```diff
@@ -10,7 +10,7 @@
 
 # This is the "version" of the whole base_deps_versions dict.
 # Change this when you change base_deps_versions.
-base_deps_meta_version = '1'
+base_deps_meta_version = '1.0.1'
```
mgautierfr: Please, keep only on number to this version number. This is just an "internal" incrementing number to be sure that travisCI get the right up-to-date archive with the prebuild dependencies.

If you increment this number, it is because the `base_deps_versions` has changed and it is not the case.
Please also update the `Gradle` entry in `base_deps_versions` (yes, the version of number of Gradle was not matching the real version used, but it is the occasion to fix this :stuck_out_tongue:  )

mgautierfr:
> I think we might be because the build had to recompile dependencies but we should probably look into using the travis wait command.

No, it is because the compilation of kiwix-android (only) takes more than 10 minutes.
I would like to see this compilation step far more faster, but I don't know how to do it :(

If we cannot reduce this compilation time, yes, we should reintroduce travis_wait. (But `travis_wait` captures all logs and prints them only at the end. It becomes difficult to monitor running build)

mhutti1:
True. It will increase kiwix-android time though but I guess it needs to be done.

mhutti1:
Why is the build now failing?

mgautierfr:
> Why is the build now failing?

It is a bug with the build of libaria2 (nothing related to your PR)

However, I've just mess-up with my git command and close by mistake your PR. I'v just recreate a new PR (#144) with your modification (but base on last version of kiwix-build)

PR #147, opened 2018-04-07
Add script to build vanilla Kiwix android app

mhutti1:


kelson42:
@mhutti1 Shoudl @mgautierfr review this merge request?

build_custom_app.py
```diff
@@ -413,7 +413,7 @@ def upload_apk(self, apk_file):
 
     def publish_release(self, options, versionCodes):
         return self.make_request('tracks', 'update',
-            track="alpha",
+            track="internal",
```
mgautierfr: Why this change ?
What is the purpose of `internal` ?
I had set the track to `alpha` to be able to "promote" the `alpha` version to `beta` and so (publish the release, ...)
Is it possible with `internal` ?

build_vanilla_app.py
```diff
@@ -0,0 +1,152 @@
...
+        return self.make_request('apks', 'upload',
+            media_body=apk_file)
+
+    def publish_release(self, options, versionCodes):
+        return self.make_request('tracks', 'update',
+            track="internal",
+            body={'versionCodes': versionCodes})
+
+
+def gen_version_code(platform_index, base_version):
```
mgautierfr: Why duplicate all those functionalities (`gen_version_code`, `Google`, `do_publish`, ...) ?
They exist already in `build_custom_app.py`, you should reuse them. (and if needed, move them in some kind of utility python module)

kelson42:
@mgautierfr The pb of all of this is that kiwix-build should does it... but it does not, so we had to do it with "our" language.

mgautierfr:
> The pb of all of this is that kiwix-build should does it... but it does not, so we had to do it with "our" language.

That's why I've opened a issue the issue #172 and said I would not block the PR for the language.
(BTW, it is not because a feature is missing that a project have to accept an implementation in any languages)

build_custom_app.py
mhutti1: Internal is a level below alpha and can be promoted to alpha beta or production. It rolls out instantly to internal testers unlike alpha which can take 3-4 hours.

kelson42:
@mhutti1 @mgautierfr What is the status on this? I this that #172 is fixed, but this is still quite unclear to me how to build easily a custom app.

kelson42:
@mhutti1 @macgills @mgautierfr We don't build anymore (from 3.0) Kiwix Android with Kiwix Build. This PR can be close.

PR #197, opened 2018-06-19
No shell

mgautierfr:


kelson42:
@mgautierfr Can we delete this PR and the branch?

mgautierfr:
No.
This is part of the job to be able to correctly run kiwix-build on windows.
I've stopped to work on this and compile kiwix-desktop using plain scripts. But at a time we will have to do the things correctly. And we will need this code.

kelson42:
@mgautierfr A ticket should be open about that (explaining the problem) and linked to this PR. Then I propose to close the PR and remove the branch using github interface. That way will easily be able to recover the code at the time needed.

mgautierfr:
We may close the PR, but why delete the branch ?

kelson42:
@mgautierfr We can do so, but i'm not really in favour of keeping branches for years.

PR #201, opened 2018-06-20
Use faster svn export instead of svn checkout

kelson42:


kelson42:
On my system, save at least 30s each time. I don't think we really need to be able to make a svn update

mgautierfr:
While this PR has been somehow superseded by #214, this change is still valuable by itself. Let's merge it.

PR #202, opened 2018-06-20
Remove support of old Xapian chert backend

kelson42:


mgautierfr:
Despite the need of update of `base_deps_meta_version`, the change in itself seems good.

mgautierfr:
No, once the PR is merged, it will be used by kiwix-build CI. 
We may push a tag on `kiwix-build` to trigger a rebuild of other projects releases, but it will just create new binaries without support of chert backend. It don't think it is necessary.

PR #203, opened 2018-06-24
Mips32r2

cm8:
Produces usable results, but for some hardcoded values (e.g. root path of toolchain) it may be desirable to supply the value on the command line, i.e. `python3 -m kiwixbuild --target-platform mips32r2_uclibc_glibcxx_dyn --toolchain-root /path/to/mips/toolchain/`.

If other mips toolchains deviate a lot in their path and binary name layout, however, then it may make more sense to create yet more (sub)classes in mips32r2.py to support these.  This will be a future choice to make for people trying builds with openwrt toolchains.

It should be safe to assume that people experienced enough to build their own mips toolchains, can find rootpath variable in mips32r2.py and re-set that accordingly.  When building with apt-get-able (debian default) mips toolchains this should not be necessary.


cm8:
depends on the value used for `--target-platform`, as you can deduct from the PR comment, you can use it to build kiwix-tools dynamically and use that on a mips target router (even if you're not allowed to install kiwix binaries and libs in the standard locations on that device).  Thus the compilation result in `INSTALL` may be used for two scenarios:
* integrate them to the root directory structure of the target device, copying suffices (the usual way)
* use them from a writable mount-point if root directory of the target is read-only/not-modifiable, copy and then call fixenv-nonstd-libdir on the target (kiwix-tools will be run through symlinks calling a shell script that sets up LD_LIBRARY_PATH in this case, for the linker to find the dynamic libraries in a non standard install location)

It may sound complicated, but use is straight-forward and its tried and tested to work.

cm8:
can you include the target `mips32r2_glibc_glibcxx_dyn` in travis ci build once it is merged to master?  it should be possible, because the toolchain used for that target is in debian and thus apt-get-able

as a side-note:  it's not necessary to rebase this branch before its merged to master (it does not have conflicts with the 3 commits in between, so it can be merged at the tip of master right away)

kelson42:
@cm8 I will let @mgautierfr make the reviewing work. We definitely wants to (1) integrate the mips architecture to the CI (2) generate statically compiled nightlies/releases for mips. If all of this works fine we should then have a look how to make it as easy as possible for openWRT users and probably optimise a few things on kiwix-serve.... But before all of this happen, please do not forget to rebase your PR branch on master.

PR #207, opened 2018-06-29
WIP: Merge origin/master in GitClone._git_update.

### LINKED ISSUE #173: Dependencies not updating?
I noticed dependencies in `/SOURCE` is not updating even if there are updates in the respective branches they are using. 

For example, I freshly cloned this repo in May 5th. Build libkiwix and dependencies. I tried to build it again on May 21. `/SOURCE/kiwix-lib` is not updated after the build. The build output tells you `gitupdate kiwix-lib : OK` But if you cd to `/SOURCE/kiwix-lib` and run `git status`, git tells you that it is xxx commits behind and can be fast forwarded. 

Looking at [`_git_update(self, context):`](https://github.com/kiwix/kiwix-build/blob/master/kiwixbuild/dependency_utils.py#L168), suppose the dependency currently use master it is gonna run:

1. git fetch origin master
2. git checkout master

I am not an expert on git. But step 2 just simply switch to local master. I think you need to do another `git merge` or simply `git pull`.

akuendig:
Otherwise, the local master branch will not contain new commits added in the origin master branch.

Fixes #173.

### DIFF
--- a/kiwixbuild/dependencies/base.py
+++ b/kiwixbuild/dependencies/base.py
@@ -158,6 +158,7 @@ class GitClone(Source):
             self.git_ref)
         run_command(command, self.git_path, context)
         run_command("git checkout "+self.git_ref, self.git_path, context)
+        run_command("git merge origin/master")
 
     def prepare(self):
         self.command('gitclone', self._git_clone)


akuendig:
Superseded by #209.

PR #210, opened 2018-06-29
Build with android logging library

mhutti1:


mgautierfr:
Sorry, I miss that PR, I was pretty busy with the build on Windows.

@mhutti1, why do you need the `log` library ?
All projects need it or just `kiwix-lib` ?
If it is just `kiwix-lib`, would it be better to add the `-llog` option to `kiwix-lib` only ?

mhutti1:
I did but kiwix-build didn't like it. I think only kiwix-lib needs it but am not sure where to add so that it works.

PR #211, opened 2018-06-29
Allow passing a message through SkipCommand.

akuendig:
This way we can log additional information to the command line in case a command is skipped for a specific reason. For example, we will be able to notify the user about the reason why a git repository was not updated.


kiwixbuild/dependencies/base.py
```diff
@@ -72,8 +72,11 @@ def command(self, name, function, *args):
             context._finalise()
             print("OK")
             return ret
-        except SkipCommand:
-            print("SKIP")
+        except SkipCommand as e:
+            message = "SKIP"
+            if e.message:
+                message += " : " + e.message
+            print(message)
```
mgautierfr: It would be better that `SkipCommand` handles the message formatting itself (using `__str__` method).
This would allow to simply have :
```
except SkipCommand as e:
    print(e)
```

akuendig: done

mgautierfr:
Superseded by #413

PR #230, opened 2018-07-31
Always use api 21 for android ndk.

mgautierfr:
mmap64 is supported only from api 21
(https://android.googlesource.com/platform/ndk/+/master/docs/user/common_problems.md)

Let's move to api 21 all the time.

### DIFF
--- a/kiwixbuild/dependencies/android_ndk.py
+++ b/kiwixbuild/dependencies/android_ndk.py
@@ -27,7 +27,7 @@ class android_ndk(Dependency):
 
         @property
         def api(self):
-            return '21' if self.arch in ('arm64', 'mips64', 'x86_64') else '14'
+            return '21'
 
         @property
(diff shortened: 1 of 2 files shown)


mhutti1:
About 5% of our users use an API version less than 21. won't this not work for them?

kelson42:
@mhutti1 @mgautierfr That seems to be a good candidate to explain https://github.com/kiwix/kiwix-android/issues/825

kelson42:
@mgautierfr @mhutti1 Do we wait something additional to merge this?

mgautierfr:
This is superseded by https://github.com/kiwix/kiwix-build/pull/231 and https://github.com/openzim/libzim/pull/178.

Once theses two PR will be merged we will use mmap when possible and fallback to "classical" memory copy/read on android 32 bits.

PR #255, opened 2018-09-14
Move to Xapian 1.4.7

kelson42:


kelson42:
@mgautierfr I think it would be good to make this move, new maintenance releases contain interesting improvments and fixes. We might also try to push kiwixbuild/patches/xapian_sys_types.patch upstream?

PR #282, opened 2019-01-08
Remove ctpp2 dependency.

mgautierfr:
kiwix-lib doesn't use ctpp2 anymore but mustache.
So remove ctpp2 dependency and add mustache.


mgautierfr:
Merging even if travis fails. This is due to the dependency changes and the fact that without ctpp2, kiwix-lib is compiled without `getHtml`, and kiwix-tools fails.
But we cannot merge the kiwix-lib's pr to use mustache because we need this PR to have mustache.

PR #283, opened 2019-01-09
Add the include dir in the meson_cross_file.

mgautierfr:
This is needed to have kiwix-lib find the `mustache.hpp` include when
cross compiling.


mgautierfr:
Same as #282, merging even if travis fails. This is due to the dependency changes and the fact that without ctpp2, kiwix-lib is compiled without getHtml, and kiwix-tools fails.
But we cannot merge the kiwix-lib's pr to use mustache because we need this PR to have mustache.

PR #288, opened 2019-02-03
Update Gradle Version

mhutti1:
Relates to https://github.com/kiwix/kiwix-android/pull/930/


mhutti1:
Ok, should have seen that. Is there anyway to make this easier? Gradle updates a lot :(

mgautierfr:
> Is there anyway to make this easier?

Not really, we will always need to update the version, the url and the sha256sum.

PR #290, opened 2019-02-04
Increase dependency version number

mhutti1:


mhutti1:
Follows on from https://github.com/kiwix/kiwix-build/pull/288

mgautierfr:
It seems that android compilation fails because licenses are not accepted.
They are not accepted because : 
- the packages id has changed 
- gradle tries to get the license for packages not installed (voluntarily)

Current installed packages are defined there : https://github.com/kiwix/kiwix-build/blob/master/kiwixbuild/dependencies/android_sdk.py#L29-L39
I succeed to compile kiwix-android by installing `tools`, `platform-tools`, `build-tools-27.0.3`, `android-27` and `android-28`.
Does the new list is good ? If yes, I will update to PR.

mhutti1:
That should be fine.

PR #306, opened 2019-03-30
Gradle Upgrade

mhutti1:
https://github.com/kiwix/kiwix-android/pull/1140

### DIFF
--- a/kiwixbuild/dependencies/gradle.py
+++ b/kiwixbuild/dependencies/gradle.py
@@ -10,9 +10,9 @@ class Gradle(Dependency):
     name = "gradle"
 
     class Source(ReleaseDownload):
-        archive = Remotefile('gradle-5.1.1-bin.zip',
-                             '4953323605c5d7b89e97d0dc7779e275bccedefcdac090aec123375eae0cc798',
-                             'https://services.gradle.org/distributions/gradle-5.1.1-bin.zip')
+        archive = Remotefile('gradle-5.2-bin.zip',
+                             'ff322863250159595e93b5a4d17a6f0d21c59a1a0497c1e1cf1d53826485503f',
+                             'https://services.gradle.org/distributions/gradle-5.2-bin.zip')
 
     class Builder(BaseBuilder):
         @property
--- a/kiwixbuild/versions.py
+++ b/kiwixbuild/versions.py
@@ -37,7 +37,6 @@ release_versions = {
 # Change this when you change base_deps_versions.
 base_deps_meta_version = '51'
 
-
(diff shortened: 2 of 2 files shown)


kelson42:
@mhutti1 What should we do with this PR?

soloturn:
@mhutti1 why there is a gradle.py to download gradle, if we are using gradle wrapper and specify it here:
https://github.com/kiwix/kiwix-android/blob/master/gradle/wrapper/gradle-wrapper.properties
?

mhutti1:
It doesn't seem necessary anymore but it was at least until recently so I am not sure.

kelson42:
@mhutti1 @soloturn I have merged this PR. Please verify if this is really needed and remove the according Python code if not.

soloturn:
@kelson42 @mhutti1 i tried to remove it but where i failed was to plug in a toolchain which does not need to be installed but is already part of the source code - the top level directory containing an executable "gradlew" script. was it @rgaudin who wrote it and could help?

rgaudin:
Nope. it looks like the gradle wrapper. Haven't been following the android build process so I can't help.

mgautierfr:
On the general PR, please read https://github.com/kiwix/kiwix-build/blob/master/kiwixbuild/versions.py#L36-L37

> why there is a gradle.py to download gradle, if we are using gradle wrapper to download gradle 

Because gradle is a build system, so a dependency. As gcc, make or meson. The wrapper is nice for people that just want to compile is one shot. But having the build system installed as a separated dependencies allow to have a better control of what we are building and how. It also allow us to install gradle with all dependencies and, in the CI, simply compile the projects instead of installing gradle at every build.

That said, if the android team what to handle everything on their side, I will be happy to remove code in kiwix-build.

soloturn:
@mgautierfr if you would be able to adjust this or show me how to would be great.

the command called nowadays is "gradlew" instead of "gradle", with exacltyl the same parameters. gradlew is a shell script in the top level of the source code, resp "gradlew.bat" on windows. this scripts are not maintained by the kiwix android team, but are kind of standard nowadays when using gradle as build system and a delivery of gradle.org:

https://docs.gradle.org/current/userguide/gradle_wrapper.html
 
gradlew downloads the version of gradle specified here:
https://github.com/kiwix/kiwix-android/tree/develop/gradle/wrapper

into a cache directory, the users home dir if nothing else is specified, and calls this downloaded gradle with the parameters passed.

soloturn:
@kelson42 who is knowledgable enough to replace gradle with gradlew ?

PR #321, opened 2019-05-07
Move from artful to bionic.

mgautierfr:
EOL for artful is july 2018.
It seems that now, apt repository has been shutdown for artful.
It is time to move to a more recent ubuntu version.

Bionic is the oldest release for which QtWebEngine has been packaged.

### DIFF
--- a/travis/appimage_builder.dockerfile
+++ b/travis/appimage_builder.dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:artful
+FROM ubuntu:bionic
 
 ENV LANG C.UTF-8
 


kelson42:
@mgautierfr I don't think they are removed, but moved to a kind of url with "archive"... in the url

mgautierfr:
Maybe. But apt doesn't work anymore.
See https://travis-ci.org/kiwix/kiwix-build/jobs/529245669

kelson42:
@mgautierfr This was just FYI, you need of course to change the source URLs in the APT configuration. But moving to bionic is great :)

PR #331, opened 2019-05-31
New android

mgautierfr:
This PR move to the last version of the ndk and build android only :

- armeabi-v7a
- arm64-v8a
- x86
- x86_64

See https://github.com/kiwix/kiwix-lib/issues/224

However, 32 bits arch (`armeabi-v7a` and `x86`) move from api 14 to 24.

According to https://android-developers.googleblog.com/2017/09/introducing-android-native-development.html, setting _FILE_OFFSET_BITS was doing nothing on old ndk version <=15, so native code was using 32bits offsets and so cannot move to offset bigger that 2Go.
With recent ndk version >15, _FILE_OFFSET_BITS now export the symbol for which 64bits offsets is available, if not, the symbol is not exported.
However, we need to move to api 24 to have all the symbol we need being exported.

I don't know how many user we lost by moving our 32bits apk from api 14 to api 24.
But, in the same time, 32 bits users probably have buggy applications for reading zim > 2Go.

cc @macgills @kelson42

### DIFF
--- a/kiwixbuild/dependencies/android_ndk.py
+++ b/kiwixbuild/dependencies/android_ndk.py
@@ -1,6 +1,6 @@
 import os
 
-from .base import Dependency, ReleaseDownload, Builder
+from .base import Dependency, ReleaseDownload, NoopBuilder
 from kiwixbuild.utils import Remotefile, add_execution_right, run_command
 
 pj = os.path.join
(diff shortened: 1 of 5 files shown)


soloturn:
current master is on api-15, see #1020. numbers from play console, may4 - jun2 2019:


```
Android 8.1 | 52,832
Android 7.1 | 31,612
Android 9 | 12,651
Android 6.0 | 12,600
Android 8.0 | 10,904
Android 7.0 | 10,776
Android 5.1 | 4,841
Android 4.4 | 4,212
Android 5.0 | 3,094
Android 4.2 | 678
Android 4.1 | 560
Android 4.3 | 507
Android 4.0.3 - 4.0.4 | 38

```

mhutti1:
I think that would mean loosing a lot of users.APi 24 is android 7

kelson42:
@mhutti1 Yes a around 3000 (from your stat) who won't be able to get the new upgrade :(

macgills:
> tl;dr: Don't set _FILE_OFFSET_BITS=64 if you want to keep the behavior present in old NDKs. 
Do we need this symbol? Over half of our installs are 32bit and if they were that buggy I really doubt the store rating would be so high.

mgautierfr:
> Do we need this symbol? Over half of our installs are 32bit and if they were that buggy I really doubt the store rating would be so high.

Ideally no, but it is the *really* common way to make 32 bits software be able to read file > 2GB. Else you would have to change all you system call from `read` to `read64` and such.
So, almost everyone define it (including our dependencies).

But anyway, this PR is superseded by https://github.com/kiwix/kiwix-build/pull/337. We are still using the old ndk. I'm closing this issue. But do not delete the branch. We will need it a moment because we will have to update the ndk.

PR #333, opened 2019-06-02
update android sdk tools to 26.1.1

soloturn:
update to sdk tools used in newest android studio 3.4.1 and gradle build tools


mgautierfr:
This is definitively not enough.
https://github.com/kiwix/kiwix-build/blob/master/kiwixbuild/dependencies/android_sdk.py#L14-L16 has be changed.

kelson42:
@mgautierfr I will rollaback this.

kelson42:
@soloturn The merge was incomplete so I have reverted it. On the top of this @macgills has implemented this dependence in gradle itself. So this won't be soon handle anyway in Kiwix-build (directly)

PR #343, opened 2019-06-27
Kiwix lib app

mgautierfr:


scripts/publish_to_bintray.py
```diff
@@ -0,0 +1,60 @@
...
+
+import requests
+from requests.auth import HTTPBasicAuth
+import os, sys
+import json
+
+bintray_auth = (os.environ['BINTRAY_USER'], os.environ['BINTRAY_PASS'])
+
+
+def create_version(version):
```
macgills: https://github.com/kiwix/kiwix-lib/blob/aa7f61c354ae6ab28829696e84e1cd7f29cf520a/android-kiwix-lib-publisher/kiwixLibAndroid/publish.gradle I am not seeing the depth of information that we used to be supplying here.
We - I think - have to be supplying the groupId at least and the artefact name.
I think the generation of a pom file is necessary? To be honest this is the first artefact I've uploaded so I don't have huge knowledge here

mgautierfr: The new version of the script also publish the pom file.

scripts/publish_to_bintray.py
```diff
@@ -0,0 +1,129 @@
...
+      <name>kiwix</name>
+      <email>contact@kiwix.org</email>
+    </developer>
+  </developers>
+  <scm>
+    <connection>https://github.com/kiwix/kiwix-lib.git</connection>
+    <developerConnection>https://github.com/kiwix/kiwix-lib.git</developerConnection>
+    <url>https://github.com/kiwix/kiwix-lib</url>
+  </scm>
+  <dependencies>
```
macgills: I don't like this dependency being here, I thought I had removed that.
Logistically the project has no dependencies I thought

mgautierfr:
Note that with this PR, the CI doesn't build kiwix-android anymore. It only build kiwix-lib for android.
This should fix https://github.com/kiwix/kiwix-android/pull/1234#issuecomment-506148260

scripts/publish_to_bintray.py
mgautierfr: This dependency is present in kiwix-lib 1.0.12.
But I can remove it if you think it is not necessary.

macgills: It was probably a mistake, should be rectified in this https://github.com/kiwix/kiwix-lib/pull/231 project which is a requirement to be merged before this is valid, should we switch context back to getting that merged?

mgautierfr: I've just removed the dependency.
(And also fix the license to GPLv3)

scripts/publish_to_bintray.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
```
macgills: maybe rename this file, it is not very generalised so "upload_kiwix_lib_android_to_bintray" might help understanding

mgautierfr: Done

kelson42:
@macgills @mgautierfr Which ticket fixes this PR? What is the status?

mgautierfr:
This is the kiwix-build side of kiwix/kiwix-lib#231

mgautierfr:
It seems there is an issue with appveyor reporting the build status.
It is ok (https://ci.appveyor.com/project/Kiwix/kiwix-build/builds/26203698), I'm merging this PR.

PR #342, opened 2019-06-26
Package kiwix-serve within kiwix-desktop.

mgautierfr:
Now that kiwix-desktop use kiwix-serve to serve the content, we need to
compile and package kiwix-serve (kiwix-tools).

Fix #341

PR #350, opened 2019-07-24
Package kiwix-serve within kiwix-desktop.

mgautierfr:
Now that kiwix-desktop use kiwix-serve to serve the content, we need to
compile and package kiwix-serve (kiwix-tools).


mgautierfr:
Supersede #342 
Use the already cross-compile kiwix-serve for windows instead of compile it.
It means that we cannot make a new release of kiwix-tools and kiwix-desktop in the same run.

mgautierfr:
Appveyor doesn't return the build status but it is ok (https://ci.appveyor.com/project/Kiwix/kiwix-build/builds/26204968)
 I'm merging.

PR #369, opened 2019-09-07
Move to Qt 5.12 everywhere

kelson42:
Fix https://github.com/kiwix/kiwix-desktop/issues/176

### DIFF
--- a/kiwixbuild/versions.py
+++ b/kiwixbuild/versions.py
@@ -42,7 +42,7 @@ release_versions = {
 
 # This is the "version" of the whole base_deps_versions dict.
 # Change this when you change base_deps_versions.
-base_deps_meta_version = '55'
+base_deps_meta_version = '56'
 
 base_deps_versions = {
   'zlib' : '1.2.8',
@@ -59,7 +59,7 @@ base_deps_versions = {
   'libmagic' : '5.35',
   'android-sdk' : 'r25.2.3',
   'android-ndk' : 'r13b',
-  'qt' : '5.10.1',
-  'qtwebengine' : '5.10.1',
+  'qt' : '5.12.4',
+  'qtwebengine' : '5.12.4',
   'org.kde' : '5.12',
 }


mgautierfr:
For Qt it is a bit tricky. We use packaged Qt in our builds :
- Installed here (https://github.com/kiwix/kiwix-build/blob/master/travis/bionic_builder.dockerfile) in the docker for linux CI container (and it is 5.9).
- Or the one installed by appveyor https://www.appveyor.com/docs/windows-images-software/#qt mainly using this line https://github.com/kiwix/kiwix-build/blob/master/appveyor/install_kiwix-desktop.cmd#L8 (so 5.11)

kelson42:
@mgautierfr Looks like Appveyor has 5.12 meanwhile... but today only 5.12.4... Here is [the ticket for 5.12.5](https://github.com/appveyor/ci/issues/3095). I guess, should be fixed soon.

kelson42:
@mgautierfr Qt 5.12.5 has been released a few days ago and Flathub has it, so problem is fixed on Flathub.

kelson42:
@mgautierfr If I understand properly, the real problem left is that the travis Docker image is on Bionic.... but I have tested, it seems to work fine with 19.10 eoan. 19.10 has still Qt 5.12.4 (see https://packages.ubuntu.com/search?keywords=qt&searchon=names) .... but 5.12.5 is already in Debian experimental (see https://packages.debian.org/search?keywords=libqt5)... so I guess should be 19.10 at the time of the release. I have uploaded an eoan version of the builder... What about moving to it? Would we then have a worse support of older computer in the appimage (I guess this is the only binary impacted)?

probonopd:
> Would we then have a worse support of older computer in the appimage

The binaries that go into AppImages should be built on a system no newer than the oldest still-supported Ubuntu LTS release. So that the resulting binary can run on all still-supported versions of Ubuntu. You can use newer Qt (from PPA) though, since a subset of Qt will be bundled alongside the application.

kelson42:
@probonopd Dou you have a specific PPA in mind?

probonopd:
https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.12.3-xenial

kelson42:
@probonopd This is not the last version of Qt 5.12 and not the proper version of Ubuntu either. We need 18.04 which is the last LTS. If we want to compile against 16.04.. then this is another ticket.

probonopd:
> This is not the last version of Qt 5.12

Which exact version do you need? I assume it will show up in Beineri's ppa soon.

> Not the proper version of Ubuntu

It is for the oldest still-supported version of Ubuntu. It is generally a good idea to build binaries on the oldest still-supported version, because the compiled application will run on _later_ versions but not on earlier versions. So, what advantage is there to building on a later version?

kelson42:
@probonopd Qt 5.12.5 has been released 10 days ago... so this PPA is many many months behind Qt's schedule.

probonopd:
Which is only really an issue in case you _need_ what was changed since. (I would not even be aware of the changes)

kelson42:
I have made Kiwix-Desktop on Windows to compile with 5.12 and it looks like @mgautierfr tries to have a more precise control of the version used on GNU/Linux appimage through custom install from Qt sources. So I close this PR.

PR #375, opened 2019-09-23
Generate the AppImage on xenial distribution.

mgautierfr:
AppImage should be genarated on a "old" distribution to be sure that
libc versions are compatible.

However, we need at least ubuntu bionic to have QWebEngine packaged.
Let's prepare our AppDir directory in bionic and generate the appimage
in xenial.
By using a small launcher script, we check that libc versions are
compatible and else print a nice message instead of crash.

See https://github.com/probonopd/linuxdeployqt/issues/340

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,16 @@ script:
       --mount=type=bind,src=$HOME/.gradle/caches,dst=/home/ci_builder/.gradle/caches \
       --mount=type=bind,src=$TRAVIS_BUILD_DIR,dst=/home/ci_builder/kiwix-build \
       kiwix/kiwix-build_ci:${VARIANT}-${DOCKER_VERSION}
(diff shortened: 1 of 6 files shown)


mgautierfr:
> Why do you think this?
launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.12.3-xenial has QWebEngine for xenial.

It could work, please see my comment here https://github.com/kiwix/kiwix-build/pull/369#issuecomment-534465755 (and continue this discussion there)

> That doesn't make it work, does it?

No, it just display a message explaining the problem (incompatible versions of glibc) instead of cryptic link error messages.

kelson42:
@mgautierfr What should we do with this PR considering that Xenial is EOL?

mgautierfr:
Nothing to do now. Now that bionic is the "older" supported distribution we are good.

PR #400, opened 2020-01-14
Use github action to build base deps of kiwix-build.

mgautierfr:
CI running with github-action use the user `runner`.
Do like github-action to be able to reuse the base_dep archive.

Use name `base_deps2_*` instead of `base_deps_*` to avoid comflict with
current CI.


mgautierfr:
The travisCI fails because of the dependencies compilation for iOs. I've got the same problem on travis and github action.
So this is a new issue and it is not related to github action. I'm merging the github action ci.

Issue #395, opened 2019-12-17
If local source code has been modified, it should not be reseted by rebuilding


kelson42:
See https://github.com/kiwix/kiwix-build/issues/346#issuecomment-566582417

Aditya-Sood:
_Gist of the issue:_ The local modifications to the code of a dependency (like kiwix-lib) will be made on a branch different from `master` of kiwix-lib. However, upon copying this local repository to the SOURCE directory of kiwix-build, the locally made changes will be ignored since the build process automatically checks-out the `master` branch of kiwix-lib.

The workaround I used was to temporarily merge the `local-changes-branch` into `master`, copy the kiwix-lib repository into kiwix-build's SOURCE directory for building kiwix-lib, and then roll back the merge with master.
Maybe some better alternative is possible which wouldn't require merging into master and then rolling-back?

PR #413, opened 2020-02-25
No git update

mgautierfr:
We simply do a `git merge --ff-only` instead of a `git reset`. This way:

- No change on the local repository
    => update to `remote/master`
- If you switch branch and local branch is behind `remote/master`
    => fast forward to `remote/master`
- If you switch branch and branches diverge
    => No update of the directory
- On branch master (or a branch behind `origin/master`) with compatible
  change in the working tree.
    =>  update of the directory
- On branch master (or a branch behind `origin/master`) but incompatible
  change in the working tree.
    =>  No update.

This way, sources are keep up-to-date as far as possible.

### DIFF
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -68,6 +68,7 @@ def run_kiwix_build(
     command = ["kiwix-build"]
     command.append(target)
     command.append("--hide-progress")
+    command.append("--fast-clone")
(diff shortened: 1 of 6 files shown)


mgautierfr:
Fixes #395 #173 
Supersede #209 #61

Issue #419, opened 2020-03-23
High level of optimization should be used


kelson42:
We have discovered that this is not the case for liblzma.

PR #420, opened 2020-03-24
Build all make based dependency using -O3 option.

mgautierfr:
This will build all Make, CMake, QMake dependency being build with -O3.
So all our dependencies will be build with -O3.

Our project (meson) are already build in release when needed.
kiwix-desktop should be build optimized in case of release but we need
to handle it correctly. At least not build in O3 per default as it may
break debug.


mgautierfr:
Fix #419

PR #433, opened 2020-04-07
Use standard name for CI

kelson42:


mgautierfr:
This is named "Build base" because this workflow build the bases (dependencies) for other projects.
We may (and will) have other workflow in the futur ("Build release", "PullRequest", ...)

kelson42:
@mgautierfr Currently what is called "Build base" in `.github/workflows/` is what is run on each PR? If not, in which file is that defined?

mgautierfr:
Yes. Ideally we should have a different workflow for the pr and the schedule build. But for now we have only one.

kelson42:
@mgautierfr AFAIK all of this is the CI (more or less directly, we just have workflows to optimize things) and we need a Workflow called "CI" like in any other repo. in our organization. I have been working to standardize the way how we do CI and how we report and I want to get it through.

PR #440, opened 2020-05-04
Gitclone for docopt dependency

MiguelRocha:
Now kiwix-build uses a commit id that has pkgconfig for docopt


kiwixbuild/dependencies/base.py
```diff
@@ -132,6 +132,7 @@ def prepare(self):
 
 class GitClone(Source):
     base_git_ref = "master"
+    git_commit = False
```
mgautierfr, line 135: I would prefer to name this `force_full_clone` or something equivalent.

mgautierfr:
I've just rebase on master and merge the two last commits.

PR #441, opened 2020-05-09
iOS remove 32bit builds

automactic:
## Remove 32bit builds for iOS. 

The Kiwix for iOS app no longer supports iOS 10, which means we no longer support 32bit devices. (All devices capable of running iOS 11 are 64 bit). I have actually already shipped two versions of the app with libs build in this branch.

I tried to update CI, but it didn't seem to work... Please help.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,9 +140,7 @@ jobs:
         target:
           - native_dyn
           - iOS_arm64
-          - iOS_i386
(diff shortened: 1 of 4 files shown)


mgautierfr:
I've remove the two ios 32bit check as mandatory check.

The fail on appveyor, is specific to windows (I'm working on it) and is not related to this PR.
I'm merging it. Thanks @automactic

PR #442, opened 2020-05-18
add vuejs as kiwix-desktop dependency

jetownfeve21:
Related to https://github.com/kiwix/kiwix-desktop/issues/230

### DIFF
--- a/kiwixbuild/dependencies/vuejs.py
+++ b/kiwixbuild/dependencies/vuejs.py
@@ -1,0 +1,37 @@
+import os
+from .base import (
+    Dependency,
+    Builder,
+    Source
+)
+from kiwixbuild.utils import Remotefile, run_command
+from kiwixbuild._global import neutralEnv
+
+pj = os.path.join
+
+class VueJs(Dependency):
+    name = "vuejs"
+
+    class Source(Source):
+        archive = Remotefile('vue.js', '', 'https://vuejs.org/js/vue.js')
+
+        def _download(self, context):
+            context.try_skip(neutralEnv('archive_dir'), self.name)
+            neutralEnv('download')(self.archive)
+
+        def prepare(self):
+            self.command('download', self._download)
+
+
+    class Builder(Builder):
+        def build(self):
+            self.command('configure', self._configure)
+
+        def make_dist(self):
+            pass
+
+        def _configure(self, context):
+            source_path = pj(neutralEnv('archive_dir'), 'vue.js')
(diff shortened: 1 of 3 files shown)


kiwixbuild/dependencies/vuejs.py
```diff
@@ -0,0 +1,37 @@
...
+from kiwixbuild.utils import Remotefile, run_command
+from kiwixbuild._global import neutralEnv
+
+pj = os.path.join
+
+class VueJs(Dependency):
+    name = "vuejs"
+
+    class Source(Source):
+        archive = Remotefile('vue.js', '', 'https://vuejs.org/js/vue.js')
```
mgautierfr, line 16: You need to file the sha256sum of the file.

kiwixbuild/dependencies/vuejs.py
```diff
@@ -0,0 +1,37 @@
...
+    class Builder(Builder):
+        def build(self):
+            self.command('configure', self._configure)
+
+        def make_dist(self):
+            pass
+
+        def _configure(self, context):
+            source_path = pj(neutralEnv('archive_dir'), 'vue.js')
+            dest_path = pj(neutralEnv('source_dir'), 'kiwix-desktop', 'resources', 'js', 'vue.js')
```
mgautierfr, line 35: I strongly dislike this. `vue.js` is a dependency of `kiwix-desktop`, it cannot change its source directory.
We could think about `kiwix-desktop` copying the file in its source but :
- It would be to kiwix-desktop to do it.
- It would be to kiwix-desktop build system to do it (not kiwix-build)
- As we build "out of source", the build system must not change the source directory. I don't know how qmake can use resource in the build directory and not in the source tree.

Honestly, If this is to complex to do (we have to check on qmake side), I would say that we keep `vue.js` in `kiwix-desktop` source tree (for now at least).

kelson42:
@jetownfeve21 Do you still plan to deal with @mgautierfr feedback?

jetownfeve21:
Sorry, I forgot about this ticket.
@mgautierfr If kiwix-build is used to download the vue.js file (or the user downloads it himself) and then qmake (kiwix-desktop) copy the file inside its sources, is it fine ?
https://wiki.qt.io/Undocumented_QMake in this link they say that the command `QMAKE_COPY` can be used (I didn't test it).

kelson42:
Problem is not that easy to fix unfortunately and we strongly consider to use native Qt in place of the HTML rendered (what would remove the vue.js). Closing for the moment.

PR #445, opened 2020-06-05
Release libzim for macOS in the CI

### LINKED ISSUE #349: Generate libzim MacOS binaries in CI
From libzim created by [ISNIT0](https://github.com/ISNIT0): openzim/libzim#203  

This is important to be able to provide easily `node-libzim` for MacOS.

See: https://github.com/openzim/libzim/issues/199

rgaudin:
Fixes #349 by enabling `native_mixed` for OSX in the CI and release.

- release on macOS is limited to libzim
- libzim shared lib is renamed to strip out the build path
- output release name is customized to produce `-macos-` instead of `-linux`

- For this to produce a macOS release, we have to change libzim version to `0` ; included it here but that may not be wanted (66eb4fa)

### DIFF
--- a/.github/scripts/build_release_nightly.py
+++ b/.github/scripts/build_release_nightly.py
@@ -13,6 +13,7 @@ from common import (
     create_desktop_image,
     update_flathub_git,
     upload_archive,
+    fix_macos_rpath,
     BASE_DIR,
     TMP_DIR,
     HOME,
@@ -35,7 +36,7 @@ elif PLATFORM_TARGET.startswith("iOS"):
     TARGETS = ("libzim", "kiwix-lib")
 elif PLATFORM_TARGET.startswith("native_"):
     if OS_NAME == "osx":
-        TARGETS = ("libzim", "zimwriterfs", "zim-tools", "kiwix-lib")
+        TARGETS = ("libzim", ) if PLATFORM_TARGET == "native_mixed" else ("libzim", "zimwriterfs", "zim-tools", "kiwix-lib")
     else:
         if DESKTOP:
             TARGETS = ("kiwix-desktop",)
@@ -59,6 +60,8 @@ for target in TARGETS:
     if target == "kiwix-desktop":
         archive = create_desktop_image(make_release=RELEASE)
     else:
+        if PLATFORM_TARGET == "native_mixed" and OS_NAME == "osx":
+            fix_macos_rpath(target)
         archive = make_archive(target, make_release=RELEASE)
     if archive:
         upload_archive(archive, target, make_release=RELEASE)
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -36,12 +36,14 @@ KIWIX_DESKTOP_ONLY = False
 _ref = _environ.get("GITHUB_REF", "").split("/")[-1]
 MAKE_RELEASE = re.fullmatch(r"r_[0-9]+", _ref) is not None
 
+RELEASE_OS_NAME = "macos" if OS_NAME == "osx" else "linux"
+
 PLATFORM_TO_RELEASE = {
-    "native_mixed": "linux-x86_64",
-    "native_static": "linux-x86_64",
+    "native_mixed": "{os}-x86_64".format(os=RELEASE_OS_NAME),
+    "native_static": "{os}-x86_64".format(os=RELEASE_OS_NAME),
     "win32_static": "win-i686",
-    "armhf_static": "linux-armhf",
-    "i586_static": "linux-i586",
+    "armhf_static": "{os}-armhf".format(os=RELEASE_OS_NAME),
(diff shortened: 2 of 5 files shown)


kelson42:
@mgautierfr @rgaudin My key criterias is that the CI pass and it does not. I suspect this is 100% unrelated to this PR but this is still a problem.

rgaudin:
> @mgautierfr @rgaudin My key criterias is that the CI pass and it does not. I suspect this is 100% unrelated to this PR but this is still a problem.

Yes it's unrelated

kelson42:
@mgautierfr Can we have a feedback on this PR please, this is quite fundamental to carry-on with other PRs on the python-libzim.

kiwixbuild/dependencies/libzim.py
```diff
@@ -29,3 +31,15 @@ def configure_option(self):
...
+
+        def _install(self, context):
+            super()._install(context)
+            platformInfo = self.buildEnv.platformInfo
+            # strip build-path info from installed libzim.dylib
+            if platformInfo.name == 'native_mixed' and option('target') == 'libzim' and neutralEnv('distname') == "Darwin":
+                for lib in pathlib.Path(self.buildEnv.install_dir).glob("**/libzim.*.dylib"):
+                    if lib.is_symlink():
+                        continue
+                    command = "install_name_tool -id {name} {path}".format(name=lib.name, path=str(lib))
```
mgautierfr, line 43: I'm not sure about this.
It should be in the libzim repository.
kiwix-build should ideally only install dependencies and launch the build system "as it should be launch". If kiwix-build has to launch `install_name_tool` it means that it is part of the normal way to compile libzim (and so, at least, being documented in libzim as the normal way to build libzim on macos).

On top of that, it seems that meson fix the rpath (and so call `install_name_tool`on mac) at installation. If this is not the case, this is maybe an bug (in libzim or meson).

kiwixbuild/versions.py
```diff
@@ -31,7 +31,7 @@
 # - set KIWIX_DESKTOP_RELEASE to 0
 
 release_versions = {
-    'libzim': None, # Depends of base deps (was 0)
+    'libzim': 0, # Depends of base deps (was None)
```
mgautierfr, line 34: I prefer to change this dictionary when we create a new release.

rgaudin: Yep, just wanted to mention that it was required to build it.

kiwixbuild/dependencies/libzim.py
rgaudin: Well that's the thing, meson does fix it at install-time, by setting it to the install's path.

```
reg@homelet:[~/src/kiwix-build]:(macos-dist)  $ otool -L BUILD_native_mixed/INSTALL/lib/libzim.6.dylib
BUILD_native_mixed/INSTALL/lib/libzim.6.dylib:
	/Users/reg/src/kiwix-build/BUILD_native_mixed/INSTALL/lib/libzim.6.dylib (compatibility version 6.0.0, current version 6.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
```

You see that the install path (`/Users/reg/src/kiwix-build/BUILD_native_mixed/INSTALL/lib/`) is present in the rpath.

meson is right in doing so as we are requesting it to install the lib.

But we want to distribute that file to users who will not install it or install it elsewhere. So this removal is a special step for this particular distribution of a *static/standalone* binary.

I am not sure about the implications of doing it in-meson.

mgautierfr: I don't know what must be done on macos.

But what you describe is a special case of the distribution of a dynamic library.
It should not be handle here, else we will remove the rpath everytime a user use kiwix-build to build libzim on macos. Either it what we want and we must move this on meson side. Or we don't and this must be moved on the distribution steps (when we create the binary archive and publish it, somewhere in `.github/scripts/common.py` or `.github/scripts/build_release_nightly.py`)

rgaudin:
@mgautierfr ok I get it.

I've moved it to build_release_nightly so it only affects the files we package for distribution/upload.
I've cleaned-up the history and removed the versions bump as well.

kelson42:
@mgautierfr Any chance to merge this today?

rgaudin:
> Only a small change about import order. Else we are good I think.

Changed.

rgaudin:
CI passed. @mgautierfr what's your strategy then to get an actual build online? Any upcoming libzim release in sight?

mgautierfr:
>  what's your strategy then to get an actual build online? Any upcoming libzim release in sight?

Now this is merged, the next nigthly should include the macos version.

I think we just can wait for the next normal release to have a official macos build.

PR #467, opened 2021-01-11
Trigger docker publish on release

### LINKED ISSUE #465: Trigger Docker build after new release/build
This is necessary to trigger a Github Action Docker build after a few new release/build:
* kiwix-serve/tools https://github.com/kiwix/kiwix-tools/issues/373
* zim-tools https://github.com/openzim/zim-tools/issues/202

rgaudin:
This triggers a `workflow_dispatch` event on the `docker.yml` workflow or the matching
repository for both `zim-tools` and `kiwix-tools` targets that supports it.

@mgautierfr this uses requests as requests is already used in upload_bintray and thus installed in the images used. I had a version using only urllib if you'd prefer.

Also, this is based on the filtering that is already done in the matrix and earlier in the script so it's just triggered after the upload of the binaries but that may not be obvious.

Finally, this just prints on error ; should we raise/fail? How critical is this given it can be triggered manually…

### DIFF
--- a/.github/scripts/build_release_nightly.py
+++ b/.github/scripts/build_release_nightly.py
@@ -14,6 +14,7 @@ from common import (
     update_flathub_git,
     upload_archive,
     fix_macos_rpath,
+    trigger_docker_publish,
     BASE_DIR,
     TMP_DIR,
     HOME,
@@ -75,6 +76,8 @@ for target in TARGETS:
         archive = make_archive(target, make_release=RELEASE)
     if archive:
         upload_archive(archive, target, make_release=RELEASE)
+        if RELEASE and target in ("zim-tools", "kiwix-tools"):
+            trigger_docker_publish(target)
 
 # We have few more things to do for release:
 if RELEASE:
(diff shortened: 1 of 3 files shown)


kelson42 (approved):

mgautierfr (changes requested):
> @mgautierfr this uses requests as requests is already used in upload_bintray and thus installed in the images used. I had a version using only urllib if you'd prefer.
> Also, this is based on the filtering that is already done in the matrix and earlier in the script so it's just triggered after the upload of the binaries but that may not be obvious.

We're good here.

> Finally, this just prints on error ; should we raise/fail? How critical is this given it can be triggered manually…

We should fail. If not we may assume that everything is fine and  we will not trigger manually. (I'm pretty sure that I will not check that docker workflow is correctly run if CI tell me no error)

(I haven't check that the url/header/json is good. I'm assuming that it is adapted to what docker workflow expects.)

rgaudin:
Thanks ; updated it accordingly. Yes json and headers are appropriate and has been manually tested.

mgautierfr:
We are good. Can you squash your commits ? (There is no need to have the history of your "wrong" attempts)

---

Appveyor is failing for something else, I'm on it.

mgautierfr (approved):

PR #468, opened 2021-01-15
Use brew ls --version

automactic:
Home brew recently deprecated `brew list`, resulting in errors when determining if automake etc. exists:

```
(venv) chrisli@Chriss-Mac-Pro kiwix-build % kiwix-build --target-platform iOS_multi kiwix-lib
[INSTALL PACKAGES]
Error: Calling `brew list` to only list formulae is disabled! Use `brew list --formula` instead.
 - autoconf : NEEDED
Error: Calling `brew list` to only list formulae is disabled! Use `brew list --formula` instead.
 - automake : NEEDED
Error: Calling `brew list` to only list formulae is disabled! Use `brew list --formula` instead.
 - libtool : NEEDED
Error: Calling `brew list` to only list formulae is disabled! Use `brew list --formula` instead.
 - cmake : NEEDED
```

In this PR, I am updating the subprocess call used to determine if a certain brew formula is installed to `brew ls --version`

### DIFF
--- a/kiwixbuild/builder.py
+++ b/kiwixbuild/builder.py
@@ -161,7 +161,7 @@ class Builder:
             package_checker = 'LANG=C dpkg -s {} 2>&1 | grep Status | grep "ok installed" 1>/dev/null 2>&1'
         elif distname == 'Darwin':
             package_installer = 'brew install {}'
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
Great

PR #470, opened 2021-02-04
zim-tools needs newly Mustache

kelson42:
See https://github.com/openzim/zim-tools/pull/206#issuecomment-772853701


veloman-yunkan (approved):
We also need @mgautierfr's assessment whether using mustache for error messages is acceptable.

mgautierfr:
> We also need @mgautierfr's assessment whether using mustache for error messages is acceptable.

I haven't review https://github.com/openzim/zim-tools/pull/206 but I agree with the idea of using mustache to format report of errors.

mgautierfr (approved):

PR #477, opened 2021-03-07
Enable inmemory backend in xapian

veloman-yunkan:
Required for kiwix/kiwix-lib#460

### DIFF
--- a/kiwixbuild/dependencies/xapian.py
+++ b/kiwixbuild/dependencies/xapian.py
@@ -21,7 +21,7 @@ class Xapian(Dependency):
         ]
 
     class Builder(MakeBuilder):
-        configure_option = "--disable-sse --disable-backend-chert --disable-backend-remote --disable-backend-inmemory --disable-documentation"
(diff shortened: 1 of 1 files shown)


kelson42:
I’m not 100% sure, but we might need to bumpup the version number to secure new CI deps are computed?!

veloman-yunkan:
> I’m not 100% sure, but we might need to bumpup the version number to secure new CI deps are computed?!

The comment makes sense, however the problem is that the version of `xapian-core` dependency hasn't changed. Is bumping `base_deps_meta_version` enough?

https://github.com/kiwix/kiwix-build/blob/b635daae68ba88304af8a710531989d71d7cc910/kiwixbuild/versions.py#L40-L42

kelson42:
@veloman-yunkan I believe so... but we need to update to latest Xapian maintenance release anyway as well...

kelson42:
I will approve this PR and create a subsequent one to upgrade to 1.4.18.

mgautierfr:
>  Is bumping base_deps_meta_version enough?

Yes it is. (would have been) (and it would have been necessary without #478)

PR #495, opened 2021-08-03
[Appveyor] Update certificate.

mgautierfr:


kelson42:
@mgautierfr One test fails, but I guess we can merge?

mgautierfr:
> @mgautierfr One test fails, but I guess we can merge?

I've fix the test and merge. And we should not have merge this without fixing it as the falling test was the only one using the certificate.

PR #499, opened 2021-10-11
macOS Builds (arm64 + x86)

automactic:
## Rationale

Adding macOS platforms -- Intel(x86) & Apple Silicon(arm64), along with a couple of other things:

- rename the base class to `ApplePlatformInfo` since it is not just about iOS anymore
- update CI to build with the newly created macOS subclasses, removing targets that I don't use (e.g.`native_dyn`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,8 @@ jobs:
           - iOS_arm64
           - iOS_x86_64
           - iOS_Mac_ABI
+          - macOS_arm64
+          - macOS_x86_64
     runs-on: macos-latest
     env:
       SSH_KEY: /tmp/id_rsa
--- a/kiwixbuild/platforms/ios.py
+++ b/kiwixbuild/platforms/ios.py
@@ -5,15 +5,15 @@ from kiwixbuild.utils import pj, xrun_find
 from .base import PlatformInfo, MetaPlatformInfo
 
 
-class iOSPlatformInfo(PlatformInfo):
+class ApplePlatformInfo(PlatformInfo):
     build = 'iOS'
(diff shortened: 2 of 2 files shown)


kelson42:
@automactic This is only for the CI or does that produces any binary/nightly?

automactic:
> Both of them fail to build as the build assumes full static while macOS requires the final bin to be dynamically linked to the System's lib.

AKIAK, macOS apps can have static libs, iOS apps can also use dynamic libs.

> disable static on final bin. This could be done in zim-tools and kiwix-tools build systems by checking if the building system is macOS and them skip the line add_global_link_arguments('-static-libstdc++', '--static', language:'cpp').

I don't have an opinion on this, as long as libkiwix and its dependencies can be static libs

> comment-out the -fembed-bitcode CFLAGS in ios.py

It is not mandatory, but I'd like to keep it on if possible (at least for libkiwix and dependencies). [More about bitcode](https://stackoverflow.com/questions/30722606/what-does-enable-bitcode-do-in-xcode-7)

kelson42:
@mgautierfr What comes next with this PR?

kelson42:
@mgautierfr This PR is important for @automactic, can we please move on?

rgaudin:
Created #511 so we can move on with this and figure out the related details in another timeline.

rgaudin (approved):

kelson42:
I have rebased on git head master. Waiting approval of @mgautierfr to merge.

mgautierfr (approved):
I've update a bit the commit messages but nothing else.

I agree with the technical change but I trust you (@rgaudin and @automactic) for the functional part.

PR #508, opened 2021-11-26
Use recent zstdlib 1.5.0

kelson42:


kelson42:
Supperseeded by  #512

PR #513, opened 2022-01-18
Do not compile libkiwix on android with werror

mgautierfr:
The java wrapper is using deprecated methods, so we cannot compile with `werror`.


mgautierfr:
The compilation error is not related to this PR. I'm merging.

PR #522, opened 2022-03-10
Fix flatpak build.

mgautierfr:
This PR improve the kiwix-build CI in several ways:
- We build the project in the CI. Before that, we were building the dependencies only on the CI and build the project in release&Nightly workflow.
- Use fedora 35 docker image to cross-compile for Windows (instead of fedora 31)
- We build the flatpak directly on the github actions' runner instead of a docker container
- Fix the flapak build by passing the option `--nointeractive` (among some other improvement options)
- Do not build zim-tools on architecture not supported (armhf, cross-compilation to macOS)


mgautierfr:
One MacOs build is broken but it is not related to this PR.

PR #524, opened 2022-03-30
Target uploads to new dropbox address

rgaudin:
SSH server which was used to receive file uploads (CI, nightly and release) has been
migrated to a new one on a different address.
Username, Key and paths are unchanged.
Most notable changes are the use of `master.download.kiwix.org` as the target in
replacement of `mirror.download.kiwix.org` (although it would still work) and the
Port to which SSH is listening on (30022 instead of 22)

### DIFF
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -168,8 +168,15 @@ def upload(file_to_upload, host, dest_path):
         print_message("No {} to upload!", file_to_upload)
         return
 
+    if ":" in host:
+        host, port = host.split(":", 1)
+    else:
+        port = "22"
+
     command = [
         "ssh",
+        "-p",
+        port,
         "-i",
         _environ.get("SSH_KEY"),
         "-o",
@@ -182,6 +189,8 @@ def upload(file_to_upload, host, dest_path):
 
     command = [
         "scp",
+        "-P",
+        port,
         "-i",
         _environ.get("SSH_KEY"),
         "-o",
@@ -195,10 +204,10 @@ def upload(file_to_upload, host, dest_path):
 
 def upload_archive(archive, project, make_release):
     if project.startswith("kiwix-") or project in ['libkiwix']:
-        host = "ci@download.kiwix.org"
+        host = "ci@master.download.kiwix.org:30022"
         dest_path = "/data/download/"
     else:
-        host = "ci@download.openzim.org"
+        host = "ci@download.openzim.org:30022"
         dest_path = "/data/openzim/"
 
     if make_release:
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -173,22 +173,31 @@ def upload(file_to_upload, host, dest_path):
     else:
         port = "22"
 
+    # sending SFTP mkdir command to the sftp interactive mode and not batch (-b) mode
+    # as the latter would exit on any mkdir error while it is most likely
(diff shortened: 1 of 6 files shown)


rgaudin:
Ah right, you're not allowed to mkdir as there's no shell. We should scp folder instead. Will add a fix

rgaudin:
Seems OK to me ; those macOS and Windows builds failed for unrelated reasons

.github/scripts/common.py
```diff
@@ -173,22 +173,9 @@ def upload(file_to_upload, host, dest_path):
...
-        "StrictHostKeyChecking=no",
-        host,
-        "mkdir -p {}".format(dest_path),
-    ]
-    print_message("Creating dest path {}", dest_path)
-    subprocess.check_call(command)
-
     command = [
         "scp",
+        "-r",
```
mgautierfr, line 178: Have you tested that ?
It seems that `-r` allow to copy recursively directory.
But here, we are copying a file into a (non existent) directory.

I'm affraid, it copies the file and rename it to the directory name (instead of creating) the directory.

rgaudin: You're right, it creates appropriate folders if you pass it a directory as source. Let me come back to you

rgaudin: Here's another that's slightly different.
I restored the mkdir step but instead of invoking an arbitrary (`mkdir -p`) command
on the remote shell via ssh, it sends `mkdir` SFTP commands.

Unfortunately, sftp isn't smart enough to automagically create the missing remote
folders so we can't just swap it all with a single SFTP command.

This has been tested and I checked that the running images (bionic, f35, focal) do
bundle the sftp binary.

`subprocess.run()` is 3.5+ which is fine given bionic has the oldest as 3.6

Let me know is this is OK with you and I'll squash it all.

.github/scripts/common.py
```diff
@@ -173,6 +173,28 @@ def upload(file_to_upload, host, dest_path):
     else:
         port = "22"
 
+    # sending SFTP mkdir command to the sftp interactive mode and not batch (-b) mode
+    # as the latter would exit on any mkdir error while it is most likely
+    # the first parts of the destination is already present and thus can't be created
+    sftp_commands = "\n".join(
+        [
+            f"mkdir {part}"
+            for part in list(reversed(Path(dest_path).parents)) + [dest_path]
```
veloman-yunkan, line 182: Why don't you take advantage of the `-p` option of `mkdir`?

rgaudin: This is not calling the gnu/bsd mkdir binary here, it's just the `mkdir` SFTP command.

mgautierfr (commented):
Seems good to me.
Can you squash the two last commits together before we merge ? The second one is not useful now.

You can rewrite the last commit with my comment, but it is mainly about coding style. Your code is good, I will not block for this.

.github/scripts/common.py
```diff
@@ -173,6 +173,28 @@ def upload(file_to_upload, host, dest_path):
...
+        ]
+    )
```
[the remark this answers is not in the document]
rgaudin: @mgautierfr assignment expr is Python 3.8+ while this has to run on 3.6…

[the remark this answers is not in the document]
rgaudin: Rebased. I did not include your elegent alternative as it would not run on bionic (Python 3.6.9)

PR #527, opened 2022-05-12
Trust codesign using the certificate

### LINKED ISSUE #519: [CI] Timeout on macos_mixed install_name_tool
At the end of the CI process, on macos mixed, we call `install_name_tool` to change the id of the library.
But somehow, the process hangout and the build is cancelled by github
See https://github.com/kiwix/kiwix-build/runs/5268088896?check_suite_focus=true

I don't know why. Maybe @automactic has an idea.

rgaudin:
When build_release_nightly calls codesign to sign libzim.7.dylib, it appears to be hanging forever.
What's most likely happening is that Keychain Access is prompting a password request without any possibility to answer, given this is running on the CI.

It's unclear whether Keychain Access wants to confirm codesign can access the certificate or if it is trying to unlock another (System) keychain to find the certificate or key.

This addresses the former.

Fixes #519, tested in [this run](https://github.com/kiwix/kiwix-build/runs/6407707440?check_suite_focus=true)

### DIFF
--- a/.github/scripts/build_projects.py
+++ b/.github/scripts/build_projects.py
@@ -9,7 +9,6 @@ from common import (
     OS_NAME,
     PLATFORM_TARGET,
     DESKTOP,
-    notarize_macos_build,
 )
 
 if PLATFORM_TARGET.startswith("android_"):
@@ -43,5 +42,4 @@ for target in TARGETS:
     else:
         if PLATFORM_TARGET == "native_mixed" and OS_NAME == "osx":
             fix_macos_rpath(target)
-            notarize_macos_build(target)
         make_archive(target, make_release=False)
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -483,6 +483,8 @@ def notarize_macos_build(project):
     for filepath in filepaths:
         subprocess.check_call(["/usr/bin/codesign", "--force", "--sign",
                                os.getenv("SIGNING_IDENTITY", "no-signing-ident"),
+                               "--keychain",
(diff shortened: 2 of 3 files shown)


rgaudin:
Failed where it should have been fixed. Please do not review 😕

rgaudin:
@mgautierfr CI is failing because of the missing secret in Pull-Request.

mgautierfr:
We mainly use it in the `Release and nightly" workflow
Its use in CI workflow is mainly here to check that we can notarize the macos build before we have to do a release/nightly.
As nightly is run regularly, I think we can remove it from CI workflow.

mgautierfr:
The nightly now don't fails now (because you've add the secret) but the signing is rejected https://github.com/kiwix/kiwix-build/runs/6462944259?check_suite_focus=true

> /Users/runner/BUILD_native_mixed/INSTALL/lib/libzim.7.dylib: rejected
source=Unnotarized Developer ID

Does this PR fixes this @rgaudin ? (I don't see why it should)

mgautierfr:
> No I didn't add any secret. I can only imagine the base image changed.

Ok. This is surprising that it is fixed just now. 

> I'd keep it for two reasons:
>
> - cleaner keychain management by specifying an absolute path and not relying on auto-discovery across different tools.
> - explicit allowance of codesign for the certificate. May save us from similar issue in the future.

Agree with that.

I've added a commit to not sign the binary in CI. If you agree with it we can merge.

rgaudin:
👍  let's do it

PR #530, opened 2022-06-09
Remove android wrapper compilation

mgautierfr:
- Remove LibkiwixAPP:
The project was used to assemble all android compilation in one multiplatform java archive.
This allow use to remove android_sdk and Gradle builder

- Be able to publish prebuild libkiwix archive on android for nightlies and release


mgautierfr:
App Veyor failing is not related to this PR

PR #551, opened 2022-11-23
Publish the CI only if we are on a dev branch.

### LINKED ISSUE #550: Under certain condition CI (workflow 'Release&Nigthly') fails even if everything is OK
If for some reason, the workflow is run twice the same day then part of the files are already uploaded in the nightly directory when the second run passes. This gives this kind of error:
```
sftp> mkdir /
Couldn't create directory: Failure
sftp> mkdir /data
Couldn't create directory: Failure
sftp> mkdir /data/openzim
Couldn't create directory: Failure
sftp> mkdir /data/openzim/nightly
Couldn't create directory: Failure
sftp> mkdir /data/openzim/nightly/2022-11-23
Couldn't create directory: Failure
--- Sending archive with command ['scp', '-rp', '-P', '30022', '-i', '/tmp/id_rsa', '-o', 'StrictHostKeyChecking=no', '/tmp/libzim_linux-x86_64-2022-11-23.tar.gz', 'ci@download.openzim.org:/data/openzim/nightly/2022-11-23'] ---
scp: /data/openzim/nightly/2022-11-23/libzim_linux-x86_64-2022-11-23.tar.gz: Permission denied
Traceback (most recent call last):
  File "kiwix-build/.github/scripts/build_release_nightly.py", line 62, in <module>
    upload_archive(archive, target, make_release=MAKE_RELEASE)
  File "/home/runner/kiwix-build/.github/scripts/common.py", line 259, in upload_archive
    upload(archive, host, dest_path)
  File "/home/runner/kiwix-build/.github/scripts/common.py", line 232, in upload
    subprocess.check_call(command)
  File "/usr/lib/python3.6/subprocess.py", line 311, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['scp', '-rp', '-P', '30022', '-i', '/tmp/id_rsa', '-o', 'StrictHostKeyChecking=no', '/tmp/libzim_linux-x86_64-2022-11-23.tar.gz', 'ci@download.openzim.org:/data/openzim/nightly/2022-11-23']' returned non-zero exit status 1.
Error: Process completed with exit code 1.
```

It seems that this is the normal behaviour to not be allow to rewrite the binaries (I certainly agree for releases, but for nightlies I have no opinion). Anyway the CI workflow should not be run if everything is OK.

This is the case for example here:
https://github.com/kiwix/kiwix-build/actions/runs/3528493896/jobs/5918659084

... even if here this is not clear why we have the error as the job seems to be run only once.

mgautierfr:
If we try to publish all the time, we will conflict with `Release&Nigthly` workflow as we would pulbish in nigthly directory.

Fix #550

### DIFF
--- a/.github/scripts/build_projects.py
+++ b/.github/scripts/build_projects.py
@@ -47,5 +47,5 @@ for target in TARGETS:
         if PLATFORM_TARGET == "native_mixed" and OS_NAME == "osx":
             fix_macos_rpath(target)
         archive = make_archive(target, make_release=False)
-    if archive:
+    if archive and DEV_BRANCH:
         upload_archive(archive, target, make_release=False, dev_branch=DEV_BRANCH)


rgaudin (approved):
This would only prevent the merge-commit from uploading but multiple-commits within a PR would still produce multiple uploads… which would work because the 0x444 mode is only applied on non-dev branches. kinda hard to follow 😉

PR #564, opened 2023-02-07
Fix the iOsMacABI target in the CI

mgautierfr:
It seems that on last version of CI envirronement, ios' clang doesnt recognize ios13.0.

Moving to ios 14.0


mgautierfr:
@automactic I'm not sure of the impact of this change.
Can you confirm we are good to move to ios 14.0 ?
Previous CI fails is https://github.com/kiwix/kiwix-build/actions/runs/4095585180/jobs/7062598418

PR #565, opened 2023-02-07
Add env container images documentation

kelson42:
This part of the features of this repository was missing a documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,22 @@ Kiwix-build.py will create several directories:
 If you want to install all those directories elsewhere, you can pass the
 `--working-dir` option to `kiwix-build`:
 
+CI/CD containerised environments
+--------------------------------
+
+Beside the `kiwix-build` command line tool, Kiwix Build proposes
+[prepared compilation environments for multiple GNU/Linux
+systems](.github/ci_images/) too.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -164,6 +164,22 @@ Kiwix-build.py will create several directories:
...
+systems](.github/ci_images/) too.
+
+These prepared environments include: compilers, binary toolchains and
+any third-party tool necessary to run the compilation toolchain. They
+don't contain pre-compiled library dependence (like Xapian, Libicu,
+Libmicrohttps, ...) for any Kiwix solution.
+
+These container images are mostly used by Kiwix itself in many of its
+CI. But they are free to download and reused, you can find them on
+[Docker Hub](https://hub.docker.com/r/kiwix/kiwix-build_ci).
```
mgautierfr, line 181: They are more than mostly used by Kiwix for its CI. They are **intended to** be used by kiwix CI **only**.

People are free to use them. But we may break it at any time without notice. No stability guaranteed at all. 

I prefer we rephrase your paragraphe to be explicit that this is intended to be internal and they use the containers at their own risk.

mgautierfr (changes requested):

PR #566, opened 2023-02-09
Move CI container building to a dedicated CI workflow

kelson42:
This PR has for purpose to improve a few things related to the CI containres:
* Use ghcr.io image repository
* Better naming/tagging of container images
* Better release mgmt based on standard release workflows
* Isolation and simplification of the container worfklow
* Fix of the container user problems (to allow later `actions/checkout` usage)
* Use new images in kiwix-build CI


ci_images/focal_builder.dockerfile
```diff
@@ -28,7 +28,7 @@ RUN apt update -q \
   && pip3 install meson pytest gcovr requests distro
 
 # Create user
-RUN useradd --create-home runner
+RUN groupadd --gid 121 runner
+RUN useradd --uid 1001 --gid 121 --create-home runner
 USER runner
-WORKDIR /home/runner
```
mgautierfr, line 33: This (and other change in the same commit) remove the WORKDIR setting.
Is it voluntary ? Why we don't need it ?

ci_images/#bionic_builder.dockerfile#
```diff
@@ -0,0 +1,45 @@
+FROM ubuntu:bionic
```
mgautierfr, line 1: I think `ci/images/#bionic_builder.dockerfile#` is added by mistake.

ci_images/focal_builder.dockerfile
kelson42: "GitHub sets the working directory path in the GITHUB_WORKSPACE environment variable. It's recommended to not use the WORKDIR instruction in your Dockerfile." https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#workdir

ci_images/#bionic_builder.dockerfile#
kelson42: Hmmm.. I don't see it https://github.com/kiwix/kiwix-build/tree/new-ci-container-build/.github/workflows. But it would be an error

mgautierfr: This is in `ci_images` directory

mgautierfr:
Rebased on master, last commit modified to not add `ci_images/#bionic_builder.dockerfile#`

kelson42:
@mgautierfr Woukd that be ready to merge? Once this is done a new release of the container should be done (we use `dev` for the moment).

PR #568, opened 2023-02-22
Remove `share/doc` and `share/man` from the deps archive.

### LINKED ISSUE #567: `/share/doc` and `/share/man` should not be distributed in the deps archives
This is unecessary and we talk here about many MB of data.

mgautierfr:
They are useless and they use precious disk space. Fix #567


kelson42 (approved):
@mgautierfr Thank you

PR #574, opened 2023-03-14
Fix creation of base dependencies archive.

mgautierfr:
The `INSTALL_DIR` was added twice. It was not a issue as we then transform the list into a set to remove duplicated.

But with `filter_install_dir` call only on one "add", the (textual) entries are not duplicated and so, not removed. So the files where add twice.

Now we correctly filter initial `INSTALL_DIR` and we remove the second add.


kelson42:
@mgautierfr Please add the ticket number it fixes

kelson42:
@mgautierfr Who is in charge of the revieq? Who should merge?

mgautierfr:
No one is in charge of the review.
I wanted to that the PR pass twice on this change to be sure that the archive is build correctly (first) and used correctly (second). This is done.

PR #581, opened 2023-03-28
Additional master to main fixes

rgaudin:
Couple of references to master in common.py which would lead to different archive upload for non-release main commits (I think) and triggering the zim-tools and kiwix-tools workflows (for docker images) on the (now missing) master branch.

Not sure exactly what actual consequences were but it ought to be fixed anyway.

### DIFF
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -455,7 +455,7 @@ def fix_macos_rpath(project):
         subprocess.check_call(command, env=os.environ)
 
 
-def trigger_workflow(repo, workflow="docker.yml", ref="master", inputs=None):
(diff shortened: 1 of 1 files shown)


kelson42:
@rgaudin ouch... will probably fix https://github.com/openzim/libzim/issues/772

mgautierfr:
> Couple of references to master in common.py which would lead to different archive upload for non-release main commits (I think) and triggering the zim-tools and kiwix-tools workflows (for docker images) on the (now missing) master branch.

That is used to publish artifact on tmp.kiwix.org/ci when we do a PR. This allow me to download the build things  to test them.
With this change, we don't upload things for main branch (and it is ok)

> @rgaudin ouch... will probably fix https://github.com/openzim/libzim/issues/772

This is probably not related

PR #588, opened 2023-04-05
Move all our CI/CD to focal instead of bionic

### LINKED ISSUE #587: Move from Ubuntu 18.04 to Ubuntu 20.04 as mnimum version supported
Ubuntu 18.04 is EOL and this is necessary anyway to fix/implement a few tickets.

mgautierfr:
Bionic is EOL.

Fix #587

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
           - android_arm64
           - android_x86
           - android_x86_64
-        image_variant: ['bionic']
+        image_variant: ['focal']
         lib_postfix: ['/x86_64-linux-gnu']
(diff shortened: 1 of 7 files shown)


kelson42:
@mgautierfr Don't forget to update the readme

mgautierfr:
There is nothing on the readme related to that. Or I miss something ?

kelson42:
First line of the README. AFAIK we know secure that our build works on Ubunuti 20.04, not 18.04 anymore.

mgautierfr:
Good catch. I've missed it as it is split in several line (and link in the middle).

> Otherwise, are we good to merge?

Yes

PR #591, opened 2023-04-12
Fix compilation on CI macOS

mgautierfr:
We don't want to compile zim-tools on macOS_arm64_dynamic.


mgautierfr:
No, it is related to CI failing : https://github.com/kiwix/kiwix-build/actions/runs/4673635963/jobs/8277076830

kelson42:
See #593

Issue #595, opened 2023-04-18
Build Linux libzim on Bionic


rgaudin:
I know you just switched from Bionic to focal; with good reasons I believe but it has some unfortunate consequence for python-libzim distribution.

Binary python packages that are not pure python are tied to a `platform` that helps package managers (`pip`) pick the appropriate file for an installation.

This notion of _platform_ is somewhat flexible. In Linux, it defaults to `linux` but can be set more specifically to announce (and should respect) libc-compatibility. There are thus `manylinux_1`, `manylinux_2010`, `manylinux_2014`, and now it's clearer with `manylinux_2_28` and the likes.

> PEP600 states that a wheel tagged manylinux_x_y shall work on any distro based on glibc>=x.y

It's recommended to use them of course as the package manager can detect an incompatible platform while using `linux`, your user could install a non-compat one and crash at runtime.

For pylibzim, we've been incorrectly setting the platform as `manylinux1` (glic 2.5) while we were building on Focal (glibc 2.31). That's bad 🫣

Now that we have a better building process, we'd like to fix that as well but it's not possible:
- libzim is built on focal (glibc 2.31)
- [manylinux project](https://github.com/pypa/manylinux) doesn't provide an [image](https://quay.io/organization/pypa) for building on 2.31. Latest they propose if 2.28

This leaves us with the following choice:
- **A**. Build for `linux`.
  - Mostly work as our users are on 2.31+
  - Prevents us from running the symbols check in the CI
  - Prevents us from automaticly testing the wheels in the CI (tests are ran in another workflow though)
- **B**. Create a manylinux-like image with 2.31+
  - It's a lot of work
- **C**. Wait for manylinux to distribute a 2.31+ image
  - No idea about when that could be.

Hence this ticket. With a bionic-built libzim, we could properly set `manylinux_2_28`.

kelson42:
OK, then lets make a special release for libzim using bionic... in addition to the normal one using focal.

rgaudin:
Same issue for macOS as well. Given we now build libzim on macOS12, we can only offer pylibzim for macOS12+.

macOS audience for pylibzim is different than Linux (mostly developers) so we could require most recent systems.
Given the main constraint being the Github Workflow runner, I propose to stick to the earliest supported macos version on github. That would be `macos-11`. `macos-10.15` is still present but deprecated and should have been removed 6m ago.

It seems this environment variable isn't sufficient in current builds as we produce libs that are 12.0.0+.

[...]

PR #606, opened 2023-05-04
Use always fix version of macOS & Ubuntu

kelson42:


.github/workflows/ci.yml
```diff
@@ -150,7 +150,7 @@ jobs:
           - macOS_arm64_static
           - macOS_arm64_mixed
           - macOS_x86_64
-    runs-on: macos-latest
+    runs-on: macos-12
```
mgautierfr, line 153: Shouldn't be macos-11 ?
See https://github.com/kiwix/kiwix-build/issues/595#issuecomment-1514849806 (and slack discussion)

kelson42: yes, bit didn't want to change without confirmation or within this PR.

PR #607, opened 2023-05-09
Better definition of what to build in the CI.

### LINKED ISSUE #573: Better describe what have to be build by kiwix-build CI
kiwix-build CI build four different kind of things:
- Dependencies for projects. Those dependencies are packaged and publish in tmp.kiwix.org for other CI to use them.
- Projects themselves. The projects are build in the CI run on PR. This compilation is mainly use to check that all projects are compiling correctly.
- Release and Nigthtly. Build every night (nightly) and on tag (release) it build all the projects and publish them.

The definition of what have to be build is defined in the three files :
- https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/compile_all_deps.py#L15-L30
- https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/build_projects.py#L16-L40
- https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/build_release_nightly.py#L25-L53

The maintenance of this "list" is pretty complexe. We should find a better way to describe what we need to build.
This depends at least on :
- What the purpose of the build (dependencies, ci, release, nightly)
- What is the target we are building to (`native_static`, `native_dyn`, `win32_*`, `iOS_*, `android_*`, `flatpak`, ...)
- On what we are building (linux, macos, windows, arm)

mgautierfr:
The current way to define what to build can be error prone as it is a succession of `if`.
Fix #573

This commit tries to fix that by introducing a table of what to build.

The definition algorithm is based on the *visual* representation of the table. It works this way:
1. Reduce the column range (Range) by looping on the header lines.
2. Once the final range is found, loop on the project lines and resolve what to build.

Few details:

1.
Each line before the `======` is a header line.
Line is divided in cells (separated by `|`) .
The first cell is a key to lookup in a context.
Other cells are "value" to match with the context. A cell's value match, we reduce the column range and go the next line. The final range is the "context range".

"values" are :
- A empty cell: We always match (equivalent to `re'.*`)
- A string starting by `re'`: What follow is a regex to make match.
- A string starting by `eval'`: What follow is a string to eval before...
- A string (or a value if `eval') which must be equal to the context.

2.
Each line after the `======` is a project line.
Line is divided in cells (separated by `|`) .
The first cell is the name of the project to build (or not). Other cells are "build definition" for the current project. We match the cells' range with the "context range". For each cell matching the "context range" we apply the "build definition". For now, "build definition" is `x` (build) or nothing.

Please note that cells can span across "previous lines" as the main information is the column range and cells are just use to reduce it.


rgaudin (commented):
OK, after looking at it more closely, I understand how it works and I find it both usable and better than current code but:

- The problem it is solving is not clearly defined
- still requires a mental effort and a usage comment seems mandatory.

First suggestion would be to drop the `DESKTOP` dimension. `native_desktop` used in the workflow seems appropriate and would reduce the number of possible combination.
Unless you added it because you envision a specific need for it?

Now on the general approach, my understanding is that we receive `OS_NAME` and `PLATFORM_TARGET` from the workflow and based on that have to decide which projects to build.

I find the [current code](https://github.com/kiwix/kiwix-build/blob/64c5f20d789e686c840a5bca3d49b38d9e5b2fd1/.github/scripts/build_projects.py#L22) easy to find out whether one specific combination should build ; as with this new proposal.
The new proposal does give an _overview_ that is not present with current code but… the need for it is unclear and more importantly, it requires computing the table in your head and given it's based on condition… you still have to trust the code does the same as your brain.

Keeping the _else_ feature brings uncertainty to this which limits greatly its value IMO. But I like explicit 😅

If the goal is to be more explicit about what we want to build, I'd suggest we use a **full list of verbose** targets.

**Sample** (doesn't account for _else_)

```py
{
    "libzim": [
        "bionic-native_mixed",
        "bionic-aarch64-mixed",
        "focal-android_arm",
        "focal-android_arm64",
        "focal-android_x86",
        "focal-android_x86_64",
        "osx-iOS_arm64",
        "osx-iOS_x86_64",
        "osx-iOS_Mac_ABI",
        "osx-macOS_arm64_mixed",
        "osx-native_mixed",
        "focal-native_mixed",
        "focal-wasm",
        "focal-armhf_mixed",
        "focal-aarch64_mixed",
    ],
    "kiwix-desktop": ["focal-native_desktop"]
}
```

Kinda frustrating that we have no list of all the possible targets and OS names…but that would definitely be redundant with workflows. I understand the windows builds doesn't use them; right?

It sure will be verbose but it's not like we have to type this multiple times. It only changes when we add/remove targets/os so not frequently.
It will serve as a documentation of all the platforms we build for (that overview we were missing) and it will be easy to tweak and toggle (just comment-out an entry).

To me, such a list would be better than both current code and this PR but this is based on my understanding of the need which may be completely off (in my defense, the PR only says “can be error prone” 😉)

rgaudin:
Ah yes I wanted to suggest to bundle the release map as well ! I skipped the dependency stuff though.

I agree (mentioned it yesterday) about reversing the array ; would be more practical but the code to parse it might get a lot more complex… maybe not worth it.

Let me know if you're settled on keeping the map and I'll take a look at the code. I still prefer the plain list but I'm fine with the table.

What's your opinion on the  _else_ feature? Don't you think it creates uncertainty?

mgautierfr:
> I agree (mentioned it yesterday) about reversing the array ; would be more practical but the code to parse it might get a lot more complex… maybe not worth it.

Not so complex after all: https://gist.github.com/mgautierfr/6a6048a7d8db96eae556e82e48784d0e

> What's your opinion on the else feature? Don't you think it creates uncertainty?

Well.. the *else* feature is not really a feature. It is *if* which is a feature 😄 
The *else* was what we have at beginning. But as builds grow and complexify, I've added *if* to ~~handle~~ remove specific case.
So the *else* is more : "If you have nothing specific to do (or not do), do what you were initially designed for: Build everything and check it compiles"

rgaudin:
Should'nt we close this ad #612 supersedes it?

Issue #585, opened 2023-03-29
Publish libzim on arm64 musl


mgautierfr:
Wanted target is :
* aarch64-linux-musl (Alpine based container for RPI)
* For testing purposes we need as well x86-64 with musl

See https://github.com/kiwix/kiwix-build/issues/549#issuecomment-1317348275

rgaudin:
Since 8.2.1, we have [`aarch64-musl`](https://download.openzim.org/release/libzim/libzim_linux-aarch64-musl-8.2.1.tar.gz) binary so this is done but we need `x86_64-musl` as well so I'll leave it open

kelson42:
@mgautierfr any update here? We kind of wait on this to release next python-libzim.

PR #611, opened 2023-05-10
Add target aarch64_musl_*

mgautierfr:
I have to mention that adding a new cross compiled target never seems so simple.


kelson42:
@mgautierfr It seems we have a few problems on 4 systems

kelson42:
@mgautierfr Only one failing ompilations left, but fail scenario seems not that straight

kelson42:
@mgautierfr Does it fixes any ticket? #585 for example? The rule is no PR without ticket because this avoids this kind of unclarity.

mgautierfr:
#585 is not fixed yet. Linux x86_64_musl is not done.

PR #612, opened 2023-05-11
Move the definition of what to build in a table.

mgautierfr:
Supersede #607

### DIFF
--- a/.github/scripts/build_definition.py
+++ b/.github/scripts/build_definition.py
@@ -7,48 +7,52 @@ import csv, io, re
 # Once a cell doesn't match, skip to the next line.
 # Once a line matches, other lines are not read, so put more specific combination first.
 # Lines composed of `-` , or `=`, or starting by `#` are ignored.
+# 'B' letter means that the project is build in the CI
+# 'd' letter means that the project's dependencies are build and published to be used by the project's CI.
+# If a cell contains both, both are done.
 BUILD_DEF = """
     | OS_NAME | DESKTOP   | PLATFORM_TARGET    | libzim | libkiwix | zim-tools | kiwix-tools | kiwix-desktop |
     =======================================================================================================
 # Bionic is a special case as we need to compile libzim on old arch for python
-    | bionic  |           |                    | b      |          |           |             |               |
+    | bionic  |           |                    | B      |          |           |             |               |
     -------------------------------------------------------------------------------------------------------
 # Osx builds, build binaries on native_dyn and native_static. On anyother things, build only the libraries
-    | osx     |           | native_dyn      |        |          | b         | b           |               |
-    | osx     |           | native_static   |        |          | b         | b           |               |
-    | osx     |           | native_mixed       | b      | b        |           |             |               |
-    | osx     |           | iOS_arm64          | b      | b        |           |             |               |
-    | osx     |           | iOS_x86_64         | b      | b        |           |             |               |
-    | osx     |           | iOS_Mac_ABI        | b      | b        |           |             |               |
-    | osx     |           | macOS_arm64_static | b      | b        |           |             |               |
-    | osx     |           | macOS_arm64_mixed  | b      | b        |           |             |               |
-    | osx     |           | macOS_x86_64       | b      | b        |           |             |               |
+    | osx     |           | native_dyn         | d      | d        | dB        | B           |               |
+    | osx     |           | native_static      |        |          | B         | B           |               |
+    | osx     |           | native_mixed       | B      | B        |           |             |               |
+    | osx     |           | iOS_arm64          | dB     | B        |           |             |               |
+    | osx     |           | iOS_x86_64         | dB     | B        |           |             |               |
+    | osx     |           | iOS_Mac_ABI        | B      | B        |           |             |               |
+    | osx     |           | macOS_arm64_static | B      | B        |           |             |               |
+    | osx     |           | macOS_arm64_mixed  | B      | B        |           |             |               |
+    | osx     |           | macOS_x86_64       | B      | B        |           |             |               |
     -------------------------------------------------------------------------------------------------------
 # Build kiwix-desktop only on specific targets
-    |         | eval'True |                    |        |          |           |             | b             |
-    |         |           | flatpak            |        |          |           |             | b             |
+    |         | eval'True |                    |        |          |           |             | B             |
+    |         |           | flatpak            |        |          |           |             | B             |
     -------------------------------------------------------------------------------------------------------
-    |         |           | native_static      |        |          | b         | b           |               |
-    |         |           | native_dyn         |        |          | b         | b           |               |
-    |         |           | native_mixed    | b      | b        |           |             |               |
-    |         |           | android_arm        | b      | b        |           |             |               |
-    |         |           | android_arm64      | b      | b        |           |             |               |
-    |         |           | android_x86        | b      | b        |           |             |               |
-    |         |           | android_x86_64     | b      | b        |           |             |               |
-    |         |           | armv6_static       |        |          | b         | b           |               |
-    |         |           | armv6_dyn          |        |          | b         | b           |               |
-    |         |           | armv6_mixed        | b      |          |           |             |               |
-    |         |           | armv8_static       |        |          | b         | b           |               |
-    |         |           | armv8_dyn          |        |          | b         | b           |               |
-    |         |           | armv8_mixed        | b      |          |           |             |               |
-    |         |           | aarch64_static     |        |          | b         | b           |               |
-    |         |           | aarch64_dyn        |        |          | b         | b           |               |
-    |         |           | aarch64_mixed      | b      |          |           |             |               |
-    |         |           | win32_static       |        |          | b         | b           |               |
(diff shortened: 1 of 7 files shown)


kelson42:
Why we have still "native" in place of "linux", the definition of platforms should better not be relative/implicit... but explicit.

mgautierfr:
> Why we have still "native" in place of "linux", the definition of platforms should better not be relative/implicit... but explicit.

Well... because this is not really a the definition of platform. "native" means native, not "linux".

There is a mix a different things here, booth in the way this is used and so in the way you understand it.

The TARGET_PLATFORM initially correspond to the `--target-platform` option passed to kiwix-build. Here, `native` means native (no cross-compilation) and other platforms mean cross-compilation to the platform. (using a platform definition which use a toolchain defined as a dependency).
If you run `kiwix-build libzim --target-platform native_dyn` on a linux computer it will build a `.so` but if you run it on a macos computer it will build a `.dylib`. This is in fact the default build if you don't pass the `--target-platform` option.

What is build on the CI is initially defined in the `.yml` files (https://github.com/kiwix/kiwix-build/blob/main/.github/workflows/ci.yml#L12-L30) and it can be read as: 
Run N jobs in parallel and in each job, run kiwix-build with this target-platform option.
So, in the CI, we don't have the definition of what to build, but how to build.

Then we have the CI script, run in each context, and this build script need to know what to build in each context. (This is what the PR is about)

Then we have https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/common.py#L50-L66 which describe what to publish and under which name.

> I still don't understand why we don't have modules defining the how to compile (like for dependencies and platforms) and a list just for the what to compile.

I'm not sure to understand what you call "how to compile":
- All kiwix-build is about how to compile : Which options to pass to each build systems.
- But kiwix-build (not the CI) is not deciding in which context you build projects
- The CI already have a definition of "in which context build" and it is the `.yml` files.
- You can somehow have this information by reading the table by column (as the table purpose is to answer what to build in each context defined in the `.yml`).

It seems you are speaking about something else, but I don't know what.

.github/scripts/build_definition.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class TableDialect(csv.Dialect):
+    delimiter = '|'
+    quoting = csv.QUOTE_NONE
+    lineterminator = "\n"
+
+def strip_array(array_str):
+    """Return a iterable of lines, skiping "decorative lines" and with all values in the line's cells stripped"""
+    for line in array_str.split('\n'):
```
rgaudin, line 57:
```suggestion
    for line in array_str.splitlines():
```

.github/scripts/common.py
```diff
@@ -17,11 +17,6 @@
 )
 
 PLATFORM_TARGET = _environ["PLATFORM_TARGET"]
-if PLATFORM_TARGET == "native_desktop":
-    PLATFORM_TARGET = "native_dyn"
-    DESKTOP = True
-else:
-    DESKTOP = False
 OS_NAME = _environ["OS_NAME"]
 HOME = Path(os.path.expanduser("~"))
```
rgaudin, line 21:
```suggestion
HOME = Path.home()
```

.github/scripts/build_definition.py
```diff
@@ -0,0 +1,105 @@
...
+            or line == "="*len(line)
+            or line == "-"*len(line)
+            or line.startswith("#")):
+            continue
+        yield "|".join(c.strip() for c in line.split('|'))
+
+def selector_match(selector, value):
+    if not selector:
+        return True
+    if selector.startswith("eval'"):
```
rgaudin, line 69: Do we still need this? Now that you removed the kiwix-desktop hack…

rgaudin (approved):
Ah, now it's clear and practical.

Can you add to the comment the meaning of `d` (build dependencies) and `b` (build)?

kelson42:
> > Why we have still "native" in place of "linux", the definition of platforms should better not be relative/implicit... but explicit.
> 
> Well... because this is not really a the definition of platform. "native" means native, not "linux".
>
> There is a mix a different things here, booth in the way this is used and so in the way you understand it.
> 
> The TARGET_PLATFORM initially correspond to the `--target-platform` option passed to kiwix-build. Here, `native` means native (no cross-compilation) and other platforms mean cross-compilation to the platform. (using a platform definition which use a toolchain defined as a dependency).

OK, but this is unecessary complicated. We should have a clear list of target platforms and not magic words like "native" which have a special meaning and are replaced on the fly.

If somehow macos and linux share compilation resources, the kiwix-build user should not have to know about that.

> If you run `kiwix-build libzim --target-platform native_dyn` on a linux computer it will build a `.so` but if you run it on a macos computer it will build a `.dylib`. This is in fact the default build if you don't pass the `--target-platform` option.

Make sense that default target platform is current one... but no magic target platform.

> What is build on the CI is initially defined in the `.yml` files (https://github.com/kiwix/kiwix-build/blob/main/.github/workflows/ci.yml#L12-L30) and it can be read as: 
> Run N jobs in parallel and in each job, run kiwix-build with this target-platform option.
> So, in the CI, we don't have the definition of what to build, but how to build.

Actually it should be there. kiwix-build should not decide this. kiwix-build should be clearly instrumented by the worflow and not decide itself what to build.

> Then we have the CI script, run in each context, and this build script need to know what to build in each context. (This is what the PR is about)

kiwix-build should what it is able to build and build each software separatly based on the instrumentation specification in the workflow.

> Then we have https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/common.py#L50-L66 which describe what to publish and under which name.

same here

> > I still don't understand why we don't have modules defining the how to compile (like for dependencies and platforms) and a list just for the what to compile.
> 
> I'm not sure to understand what you call "how to compile":
> - All kiwix-build is about how to compile : Which options to pass to each build systems.
> - But kiwix-build (not the CI) is not deciding in which context you build projects
> - The CI already have a definition of "in which context build" and it is the `.yml` files.
> - You can somehow have this information by reading the table by column (as the table purpose is to answer what to build in each context defined in the `.yml`).
>
> It seems you are speaking about something else, but I don't know what.

The what is for example "kiwix-tools for linux 64 bits with musl packaged as tarball at this location". All the rest is the "how".

It shoukd be not hidden, it should be in the worflow easily readable.

I should also be able to clearly know from kiwix-build what is possible to build given the current system... and if i request something not doable, immediatly get an error that given tbe current system, it not possible to build the requested thing.

mgautierfr:
> OK, but this is unecessary complicated.

This is why we have this PR. Trying to clean up everything because things grown organically and it is complicated to maintain.


> We should have a clear list of target platforms and not magic words like "native" which have a special meaning and are replaced on the fly.

"native" is never replaced on the fly (`desktop` was replaced by `dyn` but it is not the case with the last two commits)

> If somehow macos and linux share compilation resources, the kiwix-build user should not have to know about that.
> Make sense that default target platform is current one... but no magic target platform.

"native" is not a magic platform. It is : build with the native compiler and without passing cross-compile options.
This is the canonical, easiest build.
There is no shared resources between macos and linux. It is just NO resources, NO specific configuration.
It will be the same when kiwix-build will run on windows.

> Actually it should be there. kiwix-build should not decide this. kiwix-build should be clearly instrumented by the worflow and not decide itself what to build.

This is the case. This PR change files under `.github`. They are files used by the CI. kiwix-build itself never uses them.

> The what is for example "kiwix-tools for linux 64 bits with musl packaged as tarball at this location". All the rest is the "how".

This is already the case. The what is in the CI (split in too much different places, but in the CI):
- "linux 64 bits with musl" is defined by the job with launch (with which TARGET_PLATFORM on which ci_image)
- "kiwix-tools" is this PR
- "packaged as tarball" is in https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/common.py#L50-L66 (still the CI)
- "at this location" is https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/common.py#L247-L260

All the rest is the whole kiwix-build source code (CI excepted)


> I should also be able to clearly know from kiwix-build what is possible to build given the current system... and if i request something not doable, immediatly get an error that given tbe current system, it not possible to build the requested thing.

It is already the case at the level of the target-platform. For example for android: https://github.com/kiwix/kiwix-build/blob/main/kiwixbuild/platforms/android.py#L10
And the common code : https://github.com/kiwix/kiwix-build/blob/main/kiwixbuild/builder.py#L23-L26
It is not a error, but you have the information.

> An other remark: I don't follow the whole logic about naming either. osx does not exist anymore since 10 years for example.

It is from this : https://github.com/kiwix/kiwix-build/blob/main/.github/workflows/ci.yml#L156

mgautierfr:
All this discussion about how/what and the definition of native has not its place here. It is a PR about better define what we build on the CI, we will not change target definitions or the logic of kiwix-build here. Open a issue if you want to discuss about that.

> What do mean "b" and "db" (do i have missed here an explanation somewhere).

Added some comments about them.

> Maybe it should be fixed then

Renamed to macos

kelson42:
@mgautierfr Not sure about the exact status of this PR, but this needs to be rebased (and resolving the conflicts)

PR #622, opened 2023-05-21
Use a scp cipher which goes faster

kelson42:
Attempt to speed-up scp uploads

### DIFF
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -225,6 +225,8 @@ def upload(file_to_upload, host, dest_path):
 
     command = [
         "scp",
+        "-c",
+        "aes128-ctr",
         "-rp",
         "-P",
         port,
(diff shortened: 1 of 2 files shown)


kelson42:
@rgaudin `scp` uploads fail sporadically with `client_loop: send disconnect: Broken pipe`. This is a challenge we have worked on this the past with the Zimfarm uploader. Do you see other things we should/could do to improve the situation (if we could speed-up the upload this would welcome as well).

rgaudin:
We use the same `aes128-ctr` cipher in our [zimfarm uploader](https://github.com/openzim/zimfarm/blob/main/uploader/openzim_uploader/__init__.py).

I don't think there's much more we can do regarding speed ; and we should keep in mind that this service is receiving files from multiple ends, and serving a lot of things, all that from a mechanical disk.

Regarding the sporadic failures, I believe those are caused by the server, as previously discussed.
Here's the [sshd_config](https://github.com/openzim/zimfarm/blob/main/receiver/sshd_config) with the following interesting options

```config
#MaxAuthTries 6
#MaxSessions 10
MaxSessions 20

# Send a alive requests every 5 seconds
#ClientAliveInterval 0
ClientAliveInterval 30

# Disconnect after 10 alive requests unasnwered
#ClientAliveCountMax 3
ClientAliveCountMax 10

#MaxStartups 10:30:100
MaxStartups 100
```

Many interpretations of the [doc](https://linux.die.net/man/5/sshd_config) regarding those have been tried but it's difficult to evaluate properly because we haven't been able to willingly reproduce the issue that is mostly tied to certains conditions and a number of simultaneous clients.

PR #630, opened 2023-07-26
Extend the build definition with P(ublish), S(ource), D(ocker) and platform_name.

mgautierfr:


rgaudin (approved):
LGTM ; I'm glad to see this approach bear fruits.

One question remains though: how sure are we that all linux kiwix-tools and zim-tools will be uploaded before the trigger?

Workflow uses jobs without limiting parallelization so it's quite possible that the workflow gets triggered before all of those are uploaded.

As the docker workflows builds for all supported platform, I'm afraid this may fail.

mgautierfr:
> As the docker workflows builds for all supported platform, I'm afraid this may fail.

Ho, I missed that.

So the solution is probably to use a specific job in the workflow to trigger the workflow and add dependencies between jobs (https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#example-requiring-successful-dependent-jobs)

mgautierfr:
Last commit move docker triggering at the end of all other jobs.
I don't use a github action as we need to pass the version as input and so we need to access the python code and such.

mgautierfr:
Be careful, releaseNightly workflow is not run on PR. So CI doesn't help us to check the validity of the change.

.github/workflows/releaseNigthly.yml
```diff
@@ -236,3 +236,26 @@ jobs:
       run: $GITHUB_WORKSPACE/.github/scripts/upload_failure_logs.sh
       env:
         PLATFORM_TARGET: ${{matrix.target}}
+
+
+  Trigger_Docker:
+    needs: [Linux, Macos]
```
rgaudin, line 242: Just to be clear, we don't need macOS binaries to be the images.
I suppose you added it so that we only trigger the image is everything passed.

.github/workflows/releaseNigthly.yml
```diff
@@ -236,3 +236,26 @@ jobs:
...
+      env:
+        REP: ${{github.repository}}
+    - name: Trigger docker workflow
+      shell: bash
+      run: |
+        cd $HOME
+        kiwix-build/.github/scripts/trigger_docker_workflow.py
+      env:
+        PLATFORM_TARGET: ${{matrix.target}}
+        GITHUB_PAT: ${{secrets.GHCR_TOKEN}}
```
rgaudin, line 261: I created the corresponding secret (name cannot start with `GITHUB_`…)
```suggestion
        GITHUB_PAT: ${{secrets.DOCKER_TRIGGER_GITHUB_PAT}}
```

.github/workflows/releaseNigthly.yml
mgautierfr: It wasn't :)
I will remove it, no need to block the docker if something goes wrong on macos.

.github/workflows/releaseNigthly.yml
```diff
@@ -236,3 +236,26 @@ jobs:
       run: $GITHUB_WORKSPACE/.github/scripts/upload_failure_logs.sh
       env:
         PLATFORM_TARGET: ${{matrix.target}}
+
+
+  Trigger_Docker:
+    needs: [Linux, Macos]
+    steps:
+    - name: Checkout code
```
rgaudin, line 244: I see this step is how it is done in other job. From the code I don't understand the reason not to use the _standard_ checkout action. What is it?

.github/workflows/releaseNigthly.yml
rgaudin: `GHCR_TOKEN` is the secret containing the token to trigger the docker workflows:
- it can be removed from linux job now (the whole `GITHUB_PAT` env line)
- I moved it to `DOCKER_TRIGGER_GITHUB_PAT` so its usage is more clear.
- I will then remove the repo-level `GHCR_TOKEN` from `kiwix-build` secrets. There's a `kiwix`-wide `GHCR_TOKEN` that is different (we were overwriting it) and is used to push packages to ghcr.io

.github/workflows/releaseNigthly.yml
mgautierfr: It was probably a relic of checkout in a docker image.
This is not the case here, so I've moved it to classic github action.

PR #637, opened 2023-10-13
Update emsdk version.

mgautierfr:


kelson42:
@mgautierfr Does something stop us to merge this?

mgautierfr:
It was a test to fix https://github.com/openzim/libzim/issues/828.
I don't know how it impacts kiwix-js (@Jaifroid ?)

Jaifroid:
Well, we can build from source using EMSDK v3.1.41, and we also build the libzim.a version using 3.1.41, so it would make sense for us to be in sync on the version. It **should** work with about 90% certainty (just because there can always be something unexpected), and I guess we'd find out soon enough if it doesn't...

PR #638, opened 2023-10-19
Update Apple Platforms

automactic:
- Removed MacABI (aka UIKit on Mac, it is not needed)
- 5 targets -- iOS_arm64, iOS_simulator_x86, iOS_simulator_arm64, macOS_arm64, macOS_x86
- added `mmacosx-version-min` and `mmacosx-version-min` in various places
- update min os version to macOS 12.0 & iOS 15.0

### DIFF
--- a/.github/scripts/build_definition.py
+++ b/.github/scripts/build_definition.py
@@ -15,49 +15,48 @@ import csv, io, re
 # 'D' letter means we trigger the docker forkflow to build the docker image.
 # If a cell contains several letters, all are done.
 BUILD_DEF = """
-    | OS_NAME | PLATFORM_TARGET    | libzim | libkiwix | zim-tools | kiwix-tools | kiwix-desktop | platform_name        |
-    =====================================================================================================================
+    | OS_NAME | PLATFORM_TARGET     | libzim | libkiwix | zim-tools | kiwix-tools | kiwix-desktop | platform_name        |
+    ======================================================================================================================
 # Bionic is a special case as we need to compile libzim on old arch for python
-    | bionic  | native_mixed       | BP     |          |           |             |               | linux-x86_64-bionic  |
-    | bionic  | aarch64_mixed      | BP     |          |           |             |               | linux-aarch64-bionic |
-    --------------------------------------------------------------------------------------------------------------------
-# Osx builds, build binaries on native_dyn and native_static. On anyother things, build only the libraries
-    | macos   | native_dyn         | d      | d        | dB        | B           |               |                      |
-    | macos   | native_static      |        |          | BP        | BP          |               | macos-x86_64         |
-    | macos   | native_mixed       | BP     | BP       |           |             |               | macos-x86_64         |
-    | macos   | iOS_arm64          | dB     | B        |           |             |               |                      |
-    | macos   | iOS_x86_64         | dB     | B        |           |             |               |                      |
-    | macos   | iOS_Mac_ABI        | B      | B        |           |             |               |                      |
-    | macos   | macOS_arm64_static |        |          |           |             |               |                      |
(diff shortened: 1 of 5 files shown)


automactic:
As far as I can tell, the `kiwix-build` commands did succeed. But there are a bunch of other stuff going on after the build command, which I am not very familiar with... And something post build is failing

kelson42:
@mgautierfr Can you please have a look?

mgautierfr:
Haven't had a close look, but it missing at least a change in the [build definition](https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/build_definition.py)

(That's why https://github.com/kiwix/kiwix-build/actions/runs/6575114946/job/17861547005?pr=638)

- Files in `kiwixbuild/platforms/*` define the platforms (ie, what compiler to use, which "global" options, ...)
- Files in `kiwixbuild/dependencies/*` define how to build each project/dependency (ie, options specific to the build project)
- Jobs definition in https://github.com/kiwix/kiwix-build/tree/main/.github/workflows defines what jobs we need to launch (ie, which platform to use)
- https://github.com/kiwix/kiwix-build/blob/main/.github/scripts/build_definition.py defines what to build in each job.

mgautierfr:
This PR also show a usage of c++ feature deprecated in c++17 in zim-tools.
This is fixed in https://github.com/openzim/zim-tools/pull/376

kelson42:
I'm not in favour to abandon support of macOS for libzim without clear reason. Libzim binding should better stay compatible with older versions of macOS as long as it does not create serious problems.

.github/scripts/build_definition.py
```diff
@@ -15,49 +15,48 @@
...
+# macOS builds, build binaries on native_dyn and native_static. On another things, build only the libraries
+    | macos   | native_dyn          | d      | d        | dB        | B           |               |                      |
+    | macos   | native_static       |        |          | BP        | BP          |               | macos-x86_64         |
+    | macos   | native_mixed        | BP     | BP       |           |             |               | macos-x86_64         |
+    | macos   | iOS_arm64           | B      | B        |           |             |               |                      |
+    | macos   | iOS_simulator_x86   | B      | B        |           |             |               |                      |
+    | macos   | iOS_simulator_arm64 | B      | B        |           |             |               |                      |
+    | macos   | macOS_arm64         | B      | B        |           |             |               | macos-arm64          |
+    | macos   | macOS_x86           | B      | B        |           |             |               |                      |
```
automactic, line 32: TBH I don't think I fully understand what the differences are between `B`, `dB` and `BP` (yes I did read the keys)

Obviously I am coming from a kiwix for macOS and iOS perspective, so correct me if something should be changed

automactic:
> I'm not in favour to abandon support of macOS11 for libzim without clear reason. Libzim bindings should better stay compatible with older versions of macOS as long as it does not create serious problems.

@kelson42 Hey, are there anyone else that needs libzim on macOS 11? If I (i.e. kiwix for macOS and iOS app) is the only one, then that might be a good reason to no longer support macOS 11, since the kiwix app does not support macOS 11.

mgautierfr:
We are in https://github.com/openzim/python-libzim

kelson42:
@automactic @mgautierfr @rgaudin We are in an almost emergency situation to move forward on Apple custom apps. This is TOP PRIO that we CI+CD for libzim and libkiwix for both macOS and iOS work ASAP.

kelson42:
After discussion with @rgaudin regarding the min macOS version, we agreed:
* We will use the rule applied for Kiwix for macOS/iOS which is -2 major macOS version supported
* Which has directly consequence, that from now, min macOS supported (and on which we compile) is macOS 12

I will update the CI/CD.

kelson42:
Probably superseeded by #650.

kelson42:
Superseeded by #653

PR #650, opened 2023-11-08
Apple Platforms updates

rgaudin:
This is a redo of #638 that is supposed to be more explicit. It includes:

- now building explicitly on macos13 with Xcode 15.0.1 (for all targets)
- now targeting macOS 12+ and iOS 15+
- removed bitcode as deprecated by Xcode14
- removed `-stdlib` flag as not allowed on Xcode15
- removed `iOSMacABI` that we don't need

Note that I did not rename the targets (contrary to #638). I think it may make sense to but this should be discussed first. It's important kiwix-build maintainers understand clearly the role of each target.

### DIFF
--- a/.github/scripts/build_definition.py
+++ b/.github/scripts/build_definition.py
@@ -27,7 +27,6 @@ BUILD_DEF = """
     | macos   | native_mixed       | BP     | BP       |           |             |               | macos-x86_64         |
     | macos   | iOS_arm64          | dB     | B        |           |             |               |                      |
     | macos   | iOS_x86_64         | dB     | B        |           |             |               |                      |
-    | macos   | iOS_Mac_ABI        | B      | B        |           |             |               |                      |
     | macos   | macOS_arm64_static |        |          |           |             |               |                      |
     | macos   | macOS_arm64_mixed  | BP     | BP       |           |             |               | macos-arm64          |
     | macos   | macOS_x86_64       | B      | B        |           |             |               |                      |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,6 @@ jobs:
           - native_mixed
           - iOS_arm64
           - iOS_x86_64
-          - iOS_Mac_ABI
           - macOS_arm64_static
           - macOS_arm64_mixed
           - macOS_x86_64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,11 +152,14 @@ jobs:
           - macOS_arm64_static
           - macOS_arm64_mixed
           - macOS_x86_64
-    runs-on: macos-11
+    runs-on: macos-13
     env:
       SSH_KEY: /tmp/id_rsa
       OS_NAME: macos
     steps:
+    - name: Set Xcode version (15.0.1)
+      # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode
+      run: sudo xcode-select -s /Applications/Xcode_15.0.1.app
     - name: Checkout code
       uses: actions/checkout@v3
     - name: Setup python 3.8
--- a/.github/workflows/releaseNigthly.yml
+++ b/.github/workflows/releaseNigthly.yml
@@ -138,7 +138,7 @@ jobs:
           - native_mixed
           - macOS_arm64_static
           - macOS_arm64_mixed
-    runs-on: macos-11
+    runs-on: macos-13
     env:
       SSH_KEY: /tmp/id_rsa
       OS_NAME: macos
@@ -148,6 +148,9 @@ jobs:
       KEYCHAIN_PASSWORD: mysecretpassword
       KEYCHAIN_PROFILE: build-profile
     steps:
+    - name: Set Xcode version (15.0.1)
+      # https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode
+      run: sudo xcode-select -s /Applications/Xcode_15.0.1.app
     - name: Checkout code
       uses: actions/checkout@v3
     - name: Setup python 3.8
--- a/kiwixbuild/dependencies/libkiwix.py
+++ b/kiwixbuild/dependencies/libkiwix.py
@@ -24,8 +24,6 @@ class Libkiwix(Dependency):
             platformInfo = self.buildEnv.platformInfo
             if platformInfo.build == 'android':
                 return '-Dstatic-linkage=true -Dwerror=false'
-            if platformInfo.build == 'iOS':
-                return '-Db_bitcode=true'
             if platformInfo.name == 'flatpak':
                 return '--wrap-mode=nodownload'
             if platformInfo.mixed and option('target') == 'libkiwix':
--- a/kiwixbuild/dependencies/libzim.py
+++ b/kiwixbuild/dependencies/libzim.py
@@ -30,8 +30,6 @@ class Libzim(Dependency):
             if platformInfo.build == 'android':
                 config_options.append("-DUSE_BUFFER_HEADER=false")
                 config_options.append("-Dstatic-linkage=true")
-            if platformInfo.build == 'iOS':
-                config_options.append("-Db_bitcode=true")
             if platformInfo.mixed and option('target') == 'libzim':
                 config_options.append("-Dstatic-linkage=true")
             if platformInfo.name == "flatpak":
--- a/kiwixbuild/platforms/ios.py
+++ b/kiwixbuild/platforms/ios.py
@@ -44,14 +44,12 @@ class ApplePlatformInfo(PlatformInfo):
                 '-isysroot', self.root_path,
                 '-arch', self.arch,
                 '-target',  self.target,
-                '-stdlib=libc++'
             ],
             'extra_cflags': [
                 '-fembed-bitcode',
                 '-isysroot', self.root_path,
                 '-arch', self.arch,
                 '-target', self.target,
-                '-stdlib=libc++',
                 *('-I{}'.format(include_dir) for include_dir in self.get_include_dirs())
             ],
             'host_machine': {
@@ -87,7 +85,6 @@ class ApplePlatformInfo(PlatformInfo):
         env['CFLAGS'] = ' '.join(cflags)
         env['CXXFLAGS'] = ' '.join([
             env['CFLAGS'],
-            '-stdlib=libc++',
             '-std=c++11',
             env['CXXFLAGS'],
         ])
--- a/kiwixbuild/platforms/ios.py
+++ b/kiwixbuild/platforms/ios.py
@@ -14,6 +14,7 @@ class ApplePlatformInfo(PlatformInfo):
     target = None
     sdk_name = None
     min_iphoneos_version = None
+    min_macos_version = None
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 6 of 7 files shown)


kelson42:
Small comment: we could probably use a more recent version of the Python action and Python version for this workflow.

rgaudin:
> Small comment: we could probably use a more recent version of the Python action and Python version for this workflow.

Let's do it on a separate PR that updates all actions on all workflows of the repo?

mgautierfr:
The CI on libkiwix fails because this PR has published dependencies archive and its conflict with the configuration there. (https://github.com/kiwix/libkiwix/actions/runs/6796265893/job/18475983898)

I have made the PR #651 to avoid such overwriting project's deps. But it raise that we probably have things to change on projects' CI before merging this PR.

kelson42:
@mgautierfr #651 has been merged and CI is green here. Therefore my question: what are exactly the blockers to this PR?

mgautierfr:
> But it raise that we probably have things to change on projects' CI before merging this PR.

kelson42:
@mgautierfr Your comment is very vague, we have no clear ticket, I requested an "exact" answer. I don't get why this PR can not be merged and what has to be done now.

rgaudin:
Rebased ; deps properly sent to https://tmp.kiwix.org/ci/dev_preview/apple-xcf/ ; good to merge IMO @mgautierfr

mgautierfr:
If we look at the job log in https://github.com/kiwix/libkiwix/actions/runs/6796265893/job/18475983898, we can see that tests are failing because the test binaries cannot link with libzim (which is compile for Mac OS X 13.0) :

```
dyld: Symbol not found: __ZNKSt3__115basic_stringbufIcNS_11char_traitsIcEENS_9allocatorIcEEE3strEv
  Referenced from: /Users/runner/./BUILD_native_dyn/INSTALL/lib/libzim.8.dylib (which was built for Mac OS X 13.0)
  Expected in: /usr/lib/libc++.1.dylib
```

As this PR has published the CI dependencies for libkiwix, and on libkiwix side we still compile libkiwix for macos 11, I suppose that we have a conflict here with the macos version for which binaries are compiled. But it is just a guess. 

The PR #651 just make kiwix-build PRs not publishing CI dependencies to avoid not merged PR to break our CI. But if we merge this PR, it will goes on `main` branch and we will publish (as before) dependencies build for a macos version not expected on libkiwix (and other projects) side. So we have to update all our project's CI to build on macos 13 too.

I suspect it is the only change to do. But I can't totally be sure, so the vague comment "We have to do something on project side" (determine what is actually needed is left as an exercise for the PR creator 😉)

rgaudin:
OK understood. There's an issue as we target macOS12 and not 13. Then I'll update libkiwix CI.

kelson42 (approved):
LGTM, I would merged immediatly and adapt the CI afterwards.

mgautierfr:
>  There's an issue as we target macOS12 and not 13

Is this issue fixed ?

kelson42:
@rgaudin Concretly, this PR is blocked by which issue/PR?

rgaudin:
Can we let go of all this pressure?

kelson42:
OK, but document clearly the dependencies. On this very specific topic we have a long history of tickets and PRs hanging around for weeks. This time is over, we can not afford it anymore unfortunately as we need to deliver.

rgaudin:
Then read the discussion. @mgautierfr mentioned libkiwix not being able to link to a file apparently produced with this PR. This needs to be investigated and fixed if there's an issue. You requested an update on this after working hours and you're re-requesting one just after 8am. I don't work at night and don't intend to.

rgaudin:
OK there's not much documentation on this but here are my findings:
- `vtool` shows two values in `LC_BUILD_VERSION` loader:
  - `minos`, currently `12.0` with this code
  - `sdk` `14.0`, when using Xcode 15
- `minos` is the **minimum OS version that can run the program**
- `sdk` indicates the target OS features you want to use.
- Xcode15 does not offer use of different target SDKs anymore (use to be with version 10)
- Github runner only lists one SDK (14) per Xcode (15 in our case) which confirms line above.
- Target SDK is not a requirement per say. It means developer can conditionally use features at runtime. So not a blocker.
- I can run kiwix-serve built on mac14 (minOS 12, sdk 14) properly on macOS13.

So this all confirms that we can, indeed, **build on macOS13, targeting macOS12**, by setting the appropriate flag (as we do).

The problem with @mgautierfr's action is that dylib libzim is built using `native_dyn` target which doesn't uses the ApplePlatform described here.

I could _fix_ the issue by creating a new `macOSx64Mixed` target and build libzim for it

```py
class macOSx64Mixed(MixedMixin('macOS_x86_64'), ApplePlatformInfo):
    name = 'macOS_x86_64_mixed'
    arch = cpu = 'x86_64'
    host = 'x86_64-apple-darwin'
    target = 'x86_64-apple-macos'
    sdk_name = 'macosx'
    min_iphoneos_version = None
    min_macos_version = '12.0'
```

```sh
❯ kiwix-build --target-platform native_mixed libzim
❯ vtool -show-build ./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib
./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib:
Load command 10
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform MACOS
    minos 14.0
      sdk 14.0
   ntools 1
     tool LD
  version 1015.7
```

```sh
❯ kiwix-build --target-platform macOS_x86_64_mixed libzim
❯ vtool -show-build ./BUILD_macOS_x86_64_mixed/INSTALL/lib/libzim.8.dylib
./BUILD_macOS_x86_64_mixed/INSTALL/lib/libzim.8.dylib:
Load command 10
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform MACOS
    minos 12.0
      sdk 14.0
   ntools 1
     tool LD
  version 1015.7
```

@mgautierfr what should we do? Adding it would mean changing the CD targets for macOS

rgaudin:
Unfortunately, libzim built this way lacks a lot of symbols (and weights about half the native_mixed version)

mgautierfr:
Hum.. I'm not yet ready to drop native_platform. This is the base way to compile project and it should simply work. (Saying that, native build IS working, you can run the program on the platform you have build it. But not relevant here)

I see two options for now :

- We can set the option in the nativePlatform itself:

```patch
diff --git a/kiwixbuild/platforms/native.py b/kiwixbuild/platforms/native.py
index 8e10401..8b08da7 100644
--- a/kiwixbuild/platforms/native.py
+++ b/kiwixbuild/platforms/native.py
@@ -9,10 +9,12 @@ class NativePlatformInfo(PlatformInfo):
 
     def get_env(self):
         env = super().get_env()
         if neutralEnv('distname') == 'fedora':
             env['QT_SELECT'] = "5-64"
+        if neutralEnv('distname') == 'Darwin':
+            env['CFLAGS'] += ' '.join([env['CFLAGS'], '-mmacosx-version-min=12.0'])
         return env
 
 
 class NativeDyn(NativePlatformInfo):
     name = 'native_dyn'
```

- Or we can say it is a project's option and so, we have to set it in libzim (and other project ?):

```patch
diff --git a/kiwixbuild/dependencies/libzim.py b/kiwixbuild/dependencies/libzim.py
index 9a1fdca..b84e542 100644
--- a/kiwixbuild/dependencies/libzim.py
+++ b/kiwixbuild/dependencies/libzim.py
@@ -1,10 +1,10 @@
 from .base import (
     Dependency,
     GitClone,
     MesonBuilder)
-from kiwixbuild._global import option, get_target_step
+from kiwixbuild._global import option, get_target_step, neutralEnv
 
 class Libzim(Dependency):
     name = "libzim"
     force_build = True
 
@@ -46,5 +46,10 @@ class Libzim(Dependency):
         @property
         def library_type(self):
             if self.buildEnv.platformInfo.build == 'android':
                 return 'shared'
             return super().library_type
+
+
+        def set_env(self, env):
+            if neutralEnv('distname') == 'Darwin' or self.buildEnv.platformInfo.build == "iOS":
+                env['CFLAGS'] += ' '.join([env['CFLAGS'], '-mmacosx-version-min=12.0'])
```

mgautierfr:
> With your approach we are setting all those options twice…

My idea is to use only one of the two options.
Why we are setting twice ?
The first solution is only for native_build, so we don't set the option for other `--targetPlatform`. 
Or you are saying that we set option in two different places (in code) ?

mgautierfr:
> Another possibility is to change the minos tag after build using vtool.

Is `minos` a simple tag or it changes how it is compiled ?

rgaudin:
Yes, We are defining it in (poorly named) `ios.py` **and** in native.py. This will create discrepancy and lead to maintenance issues

rgaudin:
> > Another possibility is to change the minos tag after build using vtool.
> 
> Is `minos` a simple tag or it changes how it is compiled ?

Can be changed after compilation

```sh
❯ vtool -show-build ./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib
./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib:
Load command 10
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform MACOS
    minos 14.0
      sdk 14.0
   ntools 1
     tool LD
  version 1015.7

❯ vtool -set-version-min macos 12.0 14.0 -replace -output ./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib ./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib

❯ vtool -show-build ./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib
./BUILD_native_mixed/INSTALL/lib/libzim.8.dylib:
Load command 15
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 12.0
      sdk 14.0
```

rgaudin:
But as with your solution, it only takes care of the min-macos-version and also duplicates definition

rgaudin:
Could we tell NativeMixed to blend with corresponding platform in ios if on said platform?

rgaudin:
It's not much different ; we still have macOS related stuff in ios and this in native… but we need to move on an this can be discussed in a separate ticket and fixed later. I am cherry-picking this commit (and fixing it!)

mgautierfr:
> Could we tell NativeMixed to blend with corresponding platform in ios if on said platform?

We could maybe with this kind of patch (not tested)

```patch
diff --git a/kiwixbuild/platforms/native.py b/kiwixbuild/platforms/native.py
index 8e10401..e8fc15a 100644
--- a/kiwixbuild/platforms/native.py
+++ b/kiwixbuild/platforms/native.py
@@ -1,20 +1,23 @@
 from .base import PlatformInfo, MixedMixin
 
+import platform
 from kiwixbuild.utils import pj
 from kiwixbuild._global import option, neutralEnv
 
-
-class NativePlatformInfo(PlatformInfo):
-    build = 'native'
-
-    def get_env(self):
-        env = super().get_env()
-        if neutralEnv('distname') == 'fedora':
-            env['QT_SELECT'] = "5-64"
-        return env
-
+if platform.system() == 'Darwin':
+    from .ios import macOSx64Mixed # Which platform ? It will depend of the actual architecture x86 or arm64
+    NativePlatformInfo = macOSx64Mixed
+else:
+    class NativePlatformInfo(PlatformInfo):
+        build = 'native'
+
+        def get_env(self):
+            env = super().get_env()
+            if neutralEnv('distname') == 'fedora':
+                env['QT_SELECT'] = "5-64"
+            return env
 
 class NativeDyn(NativePlatformInfo):
     name = 'native_dyn'
     static = False
     compatible_hosts = ['fedora', 'debian', 'Darwin']
```

But it seems a bit ugly for me.

I have tried few times ago to introduce the concept of backend (ie on what we are currently running). Maybe we will need to move on this.

rgaudin:
Sorry it took so long, I wanted to make sure output is usable and realized that missing symbols were due to recent commits on libzim main (which breaks pylibzim ; will need to be updated)

rgaudin:
@mgautierfr it's good to merge IMO ; please review

PR #651, opened 2023-11-08
Publish projects deps in branch-named folder when outside of main

mgautierfr:
Still publish the deps in the dep_previews directory in case we would like to test it project's CI.

This should avoid conflict (and broken CI) when we try things on kiwix-build CI.

### DIFF
--- a/.github/scripts/compile_all_deps.py
+++ b/.github/scripts/compile_all_deps.py
@@ -6,13 +6,17 @@ from common import (
     run_kiwix_build,
     make_deps_archive,
     upload,
-    OS_NAME,
     PLATFORM_TARGET,
+    DEV_BRANCH,
 )
 from build_definition import select_build_targets, DEPS
 
 for target in select_build_targets(DEPS):
     run_kiwix_build(target, platform=PLATFORM_TARGET, build_deps_only=True)
     archive_file = make_deps_archive(target=target)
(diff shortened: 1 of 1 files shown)


rgaudin (approved):
LGTM

kelson42:
This should always have been like this and should really stay like this. Our CI pipelines are far too instable and measures should be taken and respected to fix this recurring problem.

mgautierfr:
I totally agree with you.

But, for a bit of context:

This never was like this. And it was extremely useful to check that PR on kiwix-build will actually create dependency archives which can be used on projects' CI (just have to relaunch a workflow on the project CI and see if it breaks)
If somehow the PR is broken, we just had to wait for the nightly workflow to run to reset the dep archives to the working (main) version.

We the "recent" introduction of [publishing dev preview](https://github.com/kiwix/kiwix-build/commit/a44531bb8a6bc0f0ed129465866d6e15dd35c764) in [dev_preview](https://github.com/kiwix/kiwix-build/commit/af29fd5f2e128186b669101b7f5ba4798e502579), we now have a place to also publish preview of our dependency archives.

However, to test that deps archives are ok, we now have to create a (not intended to be merged) PR on projects side with a temporary commit switching from downloading "main" dep archive to downloading custom "dev_preview" dep archive.

PR #654, opened 2023-11-09
Run command without using shell.

mgautierfr:
This is a first step to have kiwix-build running on Windows.

Windows doesn't have a shell, so running commands using `shell=True` will not work.

This is somehow a old change I've made in my attempts to move kiwix-build to Windows.
I've keep the `escape_path` stuff but I suspect some of them are missing. But we don't need it for now (no spaces in our current paths) so we will fix that when moving to Windows (with its famous 'Program Files` directory).

To be merge once we have finished the apple works, let's avoid conflicts between PRs.


Jaifroid:
> Windows doesn't have a shell

Um, PowerShell??!

kelson42:
@mgautierfr I guess we can move forward now and merge?!

PR #661, opened 2023-11-17
Release libkiwix 13.0.0

kelson42:


kiwixbuild/versions.py
```diff
@@ -31,9 +31,9 @@
 
 release_versions = {
     'libzim': None, # Depends of base deps (was None)
-    'libkiwix': None, # Depends of libzim (was 0)
+    'libkiwix': 0, # Depends of libzim (was 0)
     'kiwix-tools': None, # Depends of libkiwix and libzim (was 2)
-    'zim-tools': 0, # Depends of libzim (was None)
+    'zim-tools': None, # Depends of libzim (was None)
     'kiwix-desktop': None # Depends of libkiwix and libzim (was 4)
```
mgautierfr, line 37: We should probably also build kiwix-tools and kiwix-desktop to have new binaries using the new libkiwix.

mgautierfr (changes requested):

kiwixbuild/versions.py
kelson42: New build is probably not possible.
New release could be done for kiwix-tools, but not urgency. I will do it later.

mgautierfr: Why not possible ?

mgautierfr: Indeed, this is a major version.

PR #678, opened 2024-02-02
Build on macos-13 AND macos-14

mgautierfr:
As macos-14 is arm64, we do not publish for macos-arm64 using cross-compilation but directly use native_*


kelson42:
@mgautierfr I think what you do is not a good idea. I have not requested to change the version of macos to use to compile stuff. Macos13 compiled binaries are perfectly usable on macos14 AFAIK. We don't need two different versions of the binaries either for two diffiremt versions of macos. What we need is just properly labeled dependencies files so we take the right ones based on their respective architecture.

kelson42:
@mgautierfr What is the status of this PR?

mgautierfr:
Done with #679 #685 #689

PR #704, opened 2024-06-04
Exclude libgcc and libunwind from linked libraries on Android.

mgautierfr:
As said in [1], unwinder is linked automatically by Clang and shared library must not re-export it.

ndk after r23 (we are using r21) should do this. Else we must take care to not re-export it.

From [2], we should pass the flags `Wl,--exclude-libs,libgcc.a` and `-Wl,--exclude-libs,libunwind.a` to avoid re-export `_Unwind_Resume` methods.

Fix https://github.com/kiwix/kiwix-android/issues/3661

[1] https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#Unwinding
[2] https://github.com/android/ndk/issues/889#issuecomment-464334469

### DIFF
--- a/kiwixbuild/configs/android.py
+++ b/kiwixbuild/configs/android.py
@@ -45,7 +45,11 @@ class AndroidConfigInfo(ConfigInfo):
         return self.ndk_builder.install_path
 
     def get_cross_config(self):
-        extra_libs = ["-llog"]
+        extra_libs = [
+            "-llog",
+            "-Wl,--exclude-libs,libgcc.a",
+            "-Wl,--exclude-libs,libunwind.a",
+        ]
         extra_cflags = [
             "-I{}".format(include_dir) for include_dir in self.get_include_dirs()
         ]


kelson42:
@mgautierfr I guess it fixes the unwind error, but not the crash scenario?

mgautierfr:
@MohitMaliFtechiz, can you test with build libraries in https://tmp.kiwix.org/ci/dev_preview/fix_android_compilation please ?

> I guess it fixes the unwind error, but not the crash scenario?

The crash is due to the broken unwind. So it also fixes the crash.
A small repro case made by @MohitMaliFtechiz now don't crash on my side with this fix.

MohitMaliFtechiz:
@mgautierfr I have tested the test binary that you provided me and also with the  https://tmp.kiwix.org/ci/dev_preview/fix_android_compilation, and the application is not crashing with these binary on the arm architecture. I can read the zim file without crashing. Also, the testZim is now starts showing on the library screen so the crash scenario is fixed for the arm architecture.

Screenshot from 2024-06-05 10-48-41
Screenshot from 2024-06-05 10-48-29

mgautierfr:
Perfect. I'm merging.

PR #716, opened 2024-08-14
Use gz compression for deps archive.

mgautierfr:
It is a bit of memory waste but a lot of time win (at least on android build)

- Cold android_arm build goes from 51 minutes to 39 minutes
- Cold macos_all_static goes from 1h15 to 1h03
- Warm android_arm build goes from 32 minutes to 26 minutes
- Warm macos_all_static goes from 14 minutes to 8 minutes

### DIFF
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -343,10 +343,10 @@ def filter_install_dir(path):
                     yield sub_dir
 
 
-# Full: True if we are creating a full archive to be used as cache by kiwix-build (base_deps_{os}_{config}_{base_deps_version}.tar.xz)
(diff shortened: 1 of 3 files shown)


kelson42:
Would have been nice to update all projects as well, because this PR has broken all the CI/CD (relying on cutting-edge features of dependencies)

mgautierfr:
Arf, that is why I wanted to not have a freezed version but always use the `main` version for the `dl_deps_archive` action.

mgautierfr:
It seems only libzim is impacted, other project already use the `main` version.
PR is https://github.com/openzim/libzim/pull/911

kelson42:
It's not https://github.com/kiwix/kiwix-desktop/pull/1175, as long

PR #739, opened 2024-08-29
Do not use base_deps when building releases.

mgautierfr:
Initially we were building all our dependencies in release.
This was simply made by always compiling project using autotool or cmake
in release. As our project are using meson/qmake, all the dependencies
are build in release.

However, with recent port to Windows github CI, we have two problems:
- We have moved almost all buildsystem to meson
- On Windows, we cannot mix debug and release build.

So, when doing a release, do not use base dependencies cache.
This will extend the release workflow as we need to build everything but
at least we should build everything correctly.

### DIFF
--- a/.github/scripts/ensure_base_deps.py
+++ b/.github/scripts/ensure_base_deps.py
@@ -15,6 +15,7 @@ from common import (
     HOME,
     COMPILE_CONFIG,
     OS_NAME,
+    MAKE_RELEASE,
 )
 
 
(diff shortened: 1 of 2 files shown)


kelson42:
The risk of having CI, nightlies and releases working differently will be a potential source of instability. Considering the overall high instability of the system, I'm worried about that move and its consequences... But I'm not able to assess it properly right now.

PR #741, opened 2024-08-29
Do not build kiwix-desktop using Appveyor CI

### LINKED ISSUE #425: Build kiwix projects for Windows in the CI using github actions instead of appveyor.
The compilation of kiwix-desktop (and all dependencies) is now made on appveyor.

It is a whole process completely separated of kiwix-build and other CI.

So it is pretty complex to test on each project's CI if the PR compile on windows (on so, we don't).
We end sometime to merge code that compile on unix but not portable on window.

Now that github's action has a windows runner, we should investigate to also compile on windows using github actions.

This may not be a simple task:
- We must ensure that all c++ projects are build with the same compiler for ABI compatibilities (as we use a prebuild version of QT compile with VC++, we must use it everywhere)
- Some python tools (kiwix-build) assume a shell is available to launch subcommands)
- There are some dependencies that are use only on windows (some wrapper to provides pthreads interface). We must add it to kiwix-build (or move our projects to c++ threads library, hoping that no dependencies need pthreads).
- Compilation of windows may be "difficult" for dependencies using autotools on windows. There is mingwing/cygwin that can help here, but remember the first point.
- ...

See https://github.com/kiwix/kiwix-desktop/wiki/kiwix-desktop-windows.

### LINKED ISSUE #521: Windows CI build release from main branch
On appveyor, the CI always build on the master branch of our projects.
This is somehow ok if we build the releases just doing the releases but if it fundamentally wrong, we must build from tags.

This can be forced with specific commit as https://github.com/kiwix/kiwix-build/commit/8f31358810d69c62d852e27ab25c64a048dcc365 but it is "one shoot" commit and those kind of commit cannot be merged in master as it would force the nightly to build from the tags and not from master.

This issue would be fixed by https://github.com/kiwix/kiwix-build/issues/425 as github CI already use a a system to detect we build a release and checkout the correct code.

### LINKED ISSUE #742: [REGRESSION] Kiwix for Windows ZIP release is not correct
See [Kiwix-2024-08-30-win-amd64.zip](https://download.kiwix.org/nightly/2024-08-30/Kiwix-2024-08-30-win-amd64.zip), this is not the same like it was with Appweyor,

mgautierfr:
Fixes #521
Fixes #425
Fixes #742


mgautierfr:
kiwix-desktop windows in now published with github action CI.
And on top of that, appveyor is broken as signature certificate is expired.

PR #759, opened 2024-10-03
[TEMP] Use ci_fix branch of libkiwix

veloman-yunkan:
This PR is not intended to be merged. Its only purpose is to validate that kiwix/libkiwix#1145 actually fixes kiwix/libkiwix#1143.


veloman-yunkan:
The PR has fulfilled its function and can be closed.

PR #761, opened 2024-10-08
Enable IPv6 in meson build of libmicrohttpd under Windows

veloman-yunkan:
Should fix kiwix/libkiwix#1144

Under Windows `libmicrohttpd` was being built with support for IPv6 disabled, which - according to debugging - was the main root cause of kiwix/libkiwix#1144. This PR should eliminate that problem.

### DIFF
--- a/kiwixbuild/dependencies/libmicrohttpd.py
+++ b/kiwixbuild/dependencies/libmicrohttpd.py
@@ -21,6 +21,7 @@ class MicroHttpd(Dependency):
         patches = [
             "libmicrohttpd_meson_pkgconfig.patch",
             "libmicrohttpd_meson_timeval_tvsec_size.patch",
+            "libmicrohttpd_meson_winet6.patch",
         ]
 
     class Builder(MesonBuilder):
--- a/kiwixbuild/patches/libmicrohttpd_meson_winet6.patch
+++ b/kiwixbuild/patches/libmicrohttpd_meson_winet6.patch
@@ -1,0 +1,15 @@
+--- libmicrohttpd-0.9.76_orig/meson.build	2024-10-08 15:53:53.370828250 +0400
++++ libmicrohttpd-0.9.76/meson.build	2024-10-08 16:23:24.985668690 +0400
+@@ -77,7 +77,11 @@
+ endforeach
+
+ cdata.set('HAVE_ASSERT', cc.has_header_symbol('assert.h', 'assert'))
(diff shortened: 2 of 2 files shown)


veloman-yunkan:
To test the effect of this PR with respect on `kiwix-tools` in a preexisting build tree

1. Update the `kiwix-build` package in the virtual environment
2. Remove `libmicrohttpd` directory from under both `SOURCE` and `BUILD_native_static` directories
3. `kiwix-build --config native_static libmicrohttpd` and verify that `BUILD_native_static/libmicrohttpd/MHD_config.h` has `MHD_HAVE_INET6` `#define`d as 1
4.  `kiwix-build --config native_static kiwix-tools`
5. Test `kiwix-serve`

veloman-yunkan:
The failure of the **CI / Windows (native_static)** job looks like a glitch in a non-deterministic unit-test. I will rerun the job when the CI completes.

veloman-yunkan:
> The failure of the **CI / Windows (native_static)** job looks like a glitch in a non-deterministic unit-test. I will rerun the job when the CI completes.

Done. CI is clean.

kelson42:
@veloman-yunkan But we need to trigger the rebuild of the the dependency I guess? Maybe an opportunity to upgrade to a more recent version (not sure about how to make a simple rebuild of the same version)?

sgourdas (approved):
Works on my end as well. Nice.

veloman-yunkan:
> @veloman-yunkan But we need to trigger the rebuild of the dependency I guess?

Good point. Let me see how it can be done.

>  Maybe an opportunity to upgrade to a more recent version (not sure about how to make a simple rebuild of the same version)?

Let's not seek for trouble without any compelling reason. I will find a way to update the dependency.

kelson42:
@veloman-yunkan Maybe just updating https://github.com/kiwix/kiwix-build/blob/main/kiwixbuild/versions.py#L36 and tagging appropriatly to trigger the rebuild would be enough?

veloman-yunkan:
> @veloman-yunkan Maybe just updating https://github.com/kiwix/kiwix-build/blob/main/kiwixbuild/versions.py#L36 and tagging appropriatly to trigger the rebuild would be enough?

@kelson42 Thanks for the hint. Did it in #762. It seems that no tagging is needed. The CI of that PR will created the updated base_deps archives and after the PR is merged the new deps will be used by the nightlies.

Issue #764, opened 2024-10-13
Curl compilation relies on third party patch which whouls be hosted by us


kelson42:
Actually: https://public.kymeria.fr/KIWIX/curl_8.4.0-2_patch.zip

This should be hosted by us, in the source code of kiwix build and not as an external ZIP

kelson42:
It's now hosted by us at least thx to https://github.com/kiwix/kiwix-build/pull/765

PR #765, opened 2024-10-17
Fix missing symbols on CoreKiwix.xcframework

rgaudin:
This fixes the libcurl compilation by explicitly disabling PSL and GSS-API for https://github.com/kiwix/kiwix-apple/issues/984

Note that with this there is still an undefined symbol named  `SCDynamicStoreCopyProxies`.
This is apparently [macOS-specific](https://developer.apple.com/documentation/systemconfiguration/1517088-scdynamicstorecopyproxies) and is fixed by linking (dynamically) to `SystemConfiguration.framework`.

This doesn't seem configurable as this function is used [when `CURL_MACOS_CALL_COPYPROXIES`](https://github.com/curl/curl/blob/master/lib/macos.c) is set but this define is not exposed in config but is [set based on target](https://github.com/curl/curl/blob/master/lib/curl_setup.h#L372).

libcurl did not use/call this function in version 7.67.0 which is the one that was used prior to moving to meson.

### DIFF
--- a/kiwixbuild/dependencies/libcurl.py
+++ b/kiwixbuild/dependencies/libcurl.py
@@ -23,7 +23,7 @@ class LibCurl(Dependency):
         meson_archive = Remotefile(
             "curl_8.4.0-2_patch.zip",
             "bbb6ae75225c36ef9bb336cface729794c7c070c623a003fff40bd416042ff6e",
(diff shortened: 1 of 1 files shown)


rgaudin:
Note that this does not fix #764. It I just moved the meson ZIP to dev.kiwix

PR #771, opened 2024-10-30
Kiwix desktop jammy appimage

veloman-yunkan:
Fixes kiwix/kiwix-desktop#1219
 
Depends on kiwix/container-images#264

A working appimage was created using the updated `scripts/create_kiwix-desktop_appImage.sh` in a docker container build using the following Dockerfile  `kiwix-build_ci/jammy_builder.dockerfile` introduced by kiwix/container-images#264).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -131,12 +131,14 @@ jobs:
             image_variant: manylinux
           - config: aarch64_mixed
             image_variant: manylinux
+          - config: native_dyn
+            image_variant: jammy
     env:
       HOME: /home/runner
       SSH_KEY: /tmp/id_rsa
     runs-on: ubuntu-22.04
     container:
-      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:2024-06-03"
+      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:2024-11-30"
       options: "--device /dev/fuse --privileged"
     steps:
     - name: Checkout code
--- a/.github/scripts/build_definition.py
+++ b/.github/scripts/build_definition.py
@@ -38,32 +38,32 @@ BUILD_DEF = """
     | macos     | macOS_x86_64       | B      | B        |           |             |               |                         |                        |
     | macos     | apple_all_static   |        | BP       |           |             |               | xcframework             |                        |
     ----------------------------------------------------------------------------------------------------------------------------------------------
-    |           | flatpak            |        |          |           |             | BP            |                         |                        |
-    |           | native_static      | d      | d        | dBPSD     | dBPSD       |               | linux-x86_64            | linux-x86_64-static    |
-    |           | native_mixed       | BPS    | BPS      |           |             |               | linux-x86_64            |                        |
-    |           | native_dyn         | d      | d        | dB        | dB          | dBPS          |                         | linux-x86_64-dyn       |
+    | focal     | flatpak            |        |          |           |             | BP            |                         |                        |
+    | focal     | native_static      | d      | d        | dBPSD     | dBPSD       |               | linux-x86_64            | linux-x86_64-static    |
+    | focal     | native_mixed       | BPS    | BPS      |           |             |               | linux-x86_64            |                        |
+    | focal     | native_dyn         | d      | d        | dB        | dB          | dBPS          |                         | linux-x86_64-dyn       |
     # libzim CI is building alpine_dyn but not us
-    |           | android_arm        | dBP    | dBP      |           |             |               | android-arm             | android-arm            |
-    |           | android_arm64      | dBP    | dBP      |           |             |               | android-arm64           | android-arm64          |
-    |           | android_x86        | BP     | BP       |           |             |               | android-x86             |                        |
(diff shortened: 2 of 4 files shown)


veloman-yunkan:
So we need to introduce a `jammy` builder. The big question is which of our CI/CD jobs should we switch to that new build platform? Should we only create the `kiwix-desktop` appimage under jammy or some other builds should be upgraded as well?

kelson42:
@veloman-yunkan Great but I don't know how/what to answer to you. We probably need to move the whole linux compilation ci/cd on jammy.

veloman-yunkan:
> @veloman-yunkan Great but I don't know how/what to answer to you. We probably need to move the whole linux compilation ci/cd on jammy.

@kelson42 Then I believe that such a transition shouldn't be done as part of this PR. Rather a separate ticket must be opened and the pros and cons of the proposed upgrade should be discussed there in a broader scope.

kelson42:
@veloman-yunkan So what would be the solution for `2.4.0`?

veloman-yunkan:
> @veloman-yunkan So what would be the solution for `2.4.0`?

Maybe we can release only `kiwix-desktop` (and its appimage) from under jammy?

kelson42:
> > @veloman-yunkan So what would be the solution for `2.4.0`?
> 
> Maybe we can release only `kiwix-desktop` (and its appimage) from under jammy?

As this stage, I believe this could be a temporary solution indeed.

veloman-yunkan:
The PR is mostly ready. The most recent CI run generated an appimage built under jammy (https://tmp.kiwix.org/ci/dev_preview/kiwix-desktop_jammy_appimage/kiwix-desktop_x86_64_2024-11-25.appimage) that is devoid of the issue reported in kiwix/kiwix-desktop#1219.

The failure of the two CI runs under manylinux seems to be because of the change of the builder docker image tag from `2024-06-03` to `latest`.

veloman-yunkan:
FYI, regarding the flatpak version of kiwix-desktop - it (https://tmp.kiwix.org/ci/dev_preview/kiwix-desktop_jammy_appimage/org.kiwix.desktop.2024-11-25.flatpak which was built under focal) works fine since a newer version of Qt is used in Flatpak.

veloman-yunkan:
> The failure of the two CI runs under manylinux seems to be because of the change of the builder docker image tag from `2024-06-03` to `latest`.

I cannot find where the manylinux version of the docker image is hosted. https://ghcr.io/kiwix/kiwix-build_ci_focal works https://ghcr.io/kiwix/kiwix-build_ci_manylinux doesn't.

kelson42:
> The PR is mostly ready. The most recent CI run generated an appimage built under jammy (https://tmp.kiwix.org/ci/dev_preview/kiwix-desktop_jammy_appimage/kiwix-desktop_x86_64_2024-11-25.appimage) that is devoid of the issue reported in [kiwix/kiwix-desktop#1219](https://github.com/kiwix/kiwix-desktop/issues/1219).
> 
> The failure of the two CI runs under manylinux seems to be because of the change of the builder docker image tag from `2024-06-03` to `latest`.

Maybe installing the `requests` Python module? https://stackoverflow.com/questions/17309288/importerror-no-module-named-requests

kelson42:
> > The PR is mostly ready. The most recent CI run generated an appimage built under jammy (https://tmp.kiwix.org/ci/dev_preview/kiwix-desktop_jammy_appimage/kiwix-desktop_x86_64_2024-11-25.appimage) that is devoid of the issue reported in [kiwix/kiwix-desktop#1219](https://github.com/kiwix/kiwix-desktop/issues/1219).
> > The failure of the two CI runs under manylinux seems to be because of the change of the builder docker image tag from `2024-06-03` to `latest`.
> 
> Maybe installing the `requests` Python module? https://stackoverflow.com/questions/17309288/importerror-no-module-named-requests

@benoit74 @rgaudin Do you have an idea why the CI fails? The packages `requests` seems installed properly !!? See https://github.com/kiwix/container-images/blob/main/kiwix-build_ci/manylinux_builder.dockerfile#L16

rgaudin:
@kelson42, again, don't look at the *Upload failure logs* output. The problem is _always_ in previous steps. In this one, it's in the checkout one: `pip3: command not found`

.github/workflows/cd.yml
```diff
@@ -131,12 +131,14 @@ jobs:
...
           - config: aarch64_mixed
             image_variant: manylinux
+          - config: native_dyn
+            image_variant: jammy
     env:
       HOME: /home/runner
       SSH_KEY: /tmp/id_rsa
     runs-on: ubuntu-22.04
     container:
-      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:2024-06-03"
```
kelson42, line 139: Please come back to fixed/pinned dependency

kelson42 (changes requested):
Looks good to me and could be considered ready to merge once:
* CI is fixed, see https://github.com/kiwix/container-images/pull/265
* Docker image dependencies are pinned (again)
* @rgaudin has approved on his side

rgaudin (approved):
LGTM ; am also in favor of dedicated ticket with pros/cons for the general switch

.github/workflows/cd.yml
veloman-yunkan: Done

PR #779, opened 2024-11-22
Under Windows builds are forced to release mode

### LINKED ISSUE #757: Disable DEBUG for Windows nightly
⚠️ It's at the frontier of bug/feature.

On Windows, binaries `cl`-compiled with `_DEBUG` require the debug version of the runtime libraries (MSVCRT).

Currently, our nightlies use `_DEBUG` and thus require nightly users to install specific development tools to use them.

After discussing it with @kelson42 it appears nightlies main role is to allow power users to test the current state of projects and is not a developer/debug tool.

We should thus remove `_DEBUG` from nightlies.

What it entails is unclear at this stage ; this is in reaction of testing a kiwix-desktop. I dont know the state of shared dependencies.

On kiwix-desktop, at least there is:

```py
        @property
        def make_targets(self):
            if platform.system() == "Windows":
                yield "release-all" if option("make_release") else "debug-all"
            else:
                yield from super().make_targets

        @property
        def make_install_targets(self):
            if platform.system() == "Windows":
                yield "release-install" if option("make_release") else "debug-install"
            else:
                yield "install"
```

veloman-yunkan:
Fixes #757

`base_deps_meta_version` was bumped so that the base dependencies archives for CI workflows are recreated (we need only the Windows flavours of the archives updated but such surgical intervention is not supported).

### DIFF
--- a/kiwixbuild/dependencies/base.py
+++ b/kiwixbuild/dependencies/base.py
@@ -538,6 +538,9 @@ class MesonBuilder(Builder):
 
     @property
     def build_type(self):
+        if platform.system() == "Windows":
+            return "release"
+
         return "release" if option("make_release") else "debug"
 
     @property
--- a/kiwixbuild/dependencies/kiwix_desktop.py
+++ b/kiwixbuild/dependencies/kiwix_desktop.py
@@ -20,14 +20,14 @@ class KiwixDesktop(Dependency):
         @property
         def make_targets(self):
             if platform.system() == "Windows":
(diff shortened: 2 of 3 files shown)


veloman-yunkan:
The following fragment of README of `libzim` repo will need to be updated:

> Please notice that on Microsoft Windows with Microsoft compiler, you need to be careful to not compile in debug mode (because our released binaries are not). If you want to compile with the debug flag `/MDd`, then please use our [nightlies](https://download.openzim.org/nightly/).

PR #781, opened 2024-12-04
Fix dev.kiwix.org URL for kiwix-build's own deps

rgaudin:
`download.kiwix.org/dev/` was deprecated long ago ; this uses the correct address: https://dev.kiwix.org/kiwix-build/

- This is a mandatory fix (old URL not working anymore).
- This doesnt change anything.

Already tested in https://github.com/kiwix/kiwix-build/actions/runs/12142381918

### DIFF
--- a/kiwixbuild/utils.py
+++ b/kiwixbuild/utils.py
@@ -28,7 +28,7 @@ COLORS = {
 }
 
 
-REMOTE_PREFIX = "http://mirror.download.kiwix.org/dev/kiwix-build/"
+REMOTE_PREFIX = "https://dev.kiwix.org/kiwix-build/"
 
 
 def which(name):


kelson42 (commented):
@rgaudin My remarks (i had left a comment on the commit yesterday... but seems something went wrong):
* I strongly suspect the action should be updated as well.
* I would prefer not to mix this "cache files" with the resources needed by kiwix-build. `dev.kiwix.org/ci-cd_cache` would be a better name. Or `dev.kiwix.org/kiwix-build/ci-cd_cache` if this is important to you to have it in the `kiwix-build` directory.

rgaudin:
Again, I am **not changing anything**. The same files used to be available from both URLs:

- `[master|mirror.]download.kiwix.org/dev/` for historical reasons
- `dev.kiwix.org` the new/better address from years ago

Rebuilding the server, I removed the alias on download.kiwix.org/dev and thus changed the URL in kiwix-build to use `dev.kiwix.org`

PR #800, opened 2024-12-28
Allow to directly access ZIM file (ready-only) in Kiwix Desktop Flatpak

kelson42:
Fixes https://github.com/kiwix/kiwix-desktop/issues/1310

### DIFF
--- a/kiwixbuild/flatpak_builder.py
+++ b/kiwixbuild/flatpak_builder.py
@@ -49,6 +49,7 @@ MANIFEST = {
         "--socket=pulseaudio",
         "--share=network",
         "--share=ipc",
+        "--filesystem=host:ro",
     ],
     "cleanup": [
         "/include",


kiwixbuild/flatpak_builder.py
```diff
@@ -49,6 +49,8 @@
         "--socket=pulseaudio",
         "--share=network",
         "--share=ipc",
+        "--filesystem=host:ro",
+        ""
```
veloman-yunkan, line 53: Doesn't this extra empty argument cause any issues?

veloman-yunkan (commented):
Didn't check if it works, but the fix makes sense.

kiwixbuild/flatpak_builder.py
kelson42: I will remove this line and merge

PR #807, opened 2025-02-24
Enable SSL support in libcurl

veloman-yunkan:
Should fix kiwix/libkiwix#1180

SSL support has been disabled in our own build of `libcurl` since the very beginning.

One problem with enabling SSL is the CA certificate path that is hardcoded into `libcurl` during the configuration step. Under our Linux build system (Ubuntu) it is set to `/etc/ssl/certs/ca-certificates.crt` but on other distributions of Linux, CA certificates may reside at another location and then https URL won't work.

### DIFF
--- a/kiwixbuild/dependencies/libcurl.py
+++ b/kiwixbuild/dependencies/libcurl.py
@@ -36,7 +36,6 @@ class LibCurl(Dependency):
                 "kerberos-auth",
                 "gss-api",
                 "ssh",
-                "ssl",
                 "rtmp",
                 "http2",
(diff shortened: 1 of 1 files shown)


kelson42:
@veloman-yunkan What about other systems: Windows, iOS, macOS, Android, etc...?

veloman-yunkan:
> @veloman-yunkan What about other systems: Windows, iOS, macOS, Android, etc...?

@kelson42 I have no idea but it's likely that a similar problem may be present with them too.

PR #825, opened 2025-05-06
[TEMP] Testing the fix to libzim issue#986

veloman-yunkan:
The only purpose of this PR is to evaluate the effect of openzim/libzim#987 on kiwix-build CI before merging it.


veloman-yunkan:
This PR has served its purpose and can be closed

PR #832, opened 2025-06-06
Update to macOS 14 and XCode 16.1

### LINKED ISSUE #830: macOS CI/CD based on Xcode 16
For the moment we use Xcode 15, but:
* Xcode 16 has been released almost a year ago already (we are at version Xcode 16.4.0)
* Usually we recommend and compile on latest version of Xcode
* We can not test compilation on macOS 15 (also released almost a year ago) because our CI relies Xcode 15 (because our dependencies are compiled on Xcode 15.
* If we use Xcode 16 then we can not test on macos 13/14 anymore (but does not seem that bad) and is anyway not a strong requirement

BPerlakiH:
Fixes #830

Attempting to update to macOS 14 and XCode 16.1 on the CI/CD pipelines.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -233,7 +233,7 @@ jobs:
           - macOS_arm64_static
           - macOS_arm64_mixed
           - apple_all_static
-    runs-on: macos-13
+    runs-on: macos-14
     env:
       SSH_KEY: /tmp/id_rsa
       OS_NAME: macos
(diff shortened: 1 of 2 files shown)


kelson42:
@BPerlakiH ais that an attempt to fix #830?

kelson42 (approved):
@rgaudin LGTM, but I guess we should do anew build?

rgaudin (approved):
LGTM

PR #834, opened 2025-06-07
Use Ubuntu 22.04 (in place of 20.04) for CI/CD

kelson42:
* Use Ubuntu 22.04 in place of 20.04 in CI/CD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
           - android_arm64
           - android_x86
           - android_x86_64
-        image_variant: ['focal']
+        image_variant: ['jammy']
         include:
           - config: native_mixed
             image_variant: manylinux
@@ -118,7 +118,7 @@ jobs:
       OS_NAME: ${{matrix.image_variant}}
     runs-on: ubuntu-22.04
     container:
-      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:2024-11-30"
+      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:2025-06-06"
       options: "--device /dev/fuse --privileged"
     steps:
     - name: Checkout code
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan So I think I have brought this PR as far as I could with my knowledge. Do my understanding we face here two different kinds of problems:
* curl does not compile anymore (and we have [two serious issues still open around it](https://github.com/kiwix/kiwix-build/issues?q=is%3Aissue%20state%3Aopen%20curl)). Maybe an opportunity here to nail it down.
* It looks like the [musl binaries](https://more.musl.cc/) seem outdated and we don't have new one, so we probably need to rethink how do deal with our musl based libzim

What do you think?

veloman-yunkan:
The container images [kiwix-build_ci_jammy:2025-01-05](https://github.com/orgs/kiwix/packages/container/kiwix-build_ci_jammy/331214710?tag=2025-01-05) and [kiwix-build_ci_jammy:2025-06-06](https://github.com/orgs/kiwix/packages/container/kiwix-build_ci_jammy/432664058?tag=2025-06-06) differ in the version of `meson` - it was upgraded from 1.6.1 to 1.8.1. I believe that the easy fix is to force `meson` in the builder images to v1.6.1.

veloman-yunkan:
> The container images [kiwix-build_ci_jammy:2025-01-05](https://github.com/orgs/kiwix/packages/container/kiwix-build_ci_jammy/331214710?tag=2025-01-05) and [kiwix-build_ci_jammy:2025-06-06](https://github.com/orgs/kiwix/packages/container/kiwix-build_ci_jammy/432664058?tag=2025-06-06) differ in the version of `meson` - it was upgraded from 1.6.1 to 1.8.1. I believe that the easy fix is to force `meson` in the builder images to v1.6.1.

Done in kiwix/container-images#280

kelson42:
@veloman-yunkan Nice, I believe you have fixed the curl problem leave musl to be the last one.

kelson42:
@veloman-yunkan This is acceptable

.github/workflows/ci.yml
```diff
@@ -112,6 +110,10 @@ jobs:
             image_variant: manylinux
           - config: native_dyn
             image_variant: jammy
```
veloman-yunkan, line 112: Same here

veloman-yunkan (commented):
LGTM, save for a small cleanup step.

PR #838, opened 2025-06-15
Windows 2025 ci/cd

kelson42:


kelson42:
@veloman-yunkan Maybe we should also update the base-deps?

kelson42:
@veloman-yunkan Et voilà ! Surprisingly everything just went smooth!

Issue #823, opened 2025-04-20
Remove Ubuntu 20.04 Focal from CI/CD


kelson42:
Reasons are:
* Times move on and 20.04 is almost out of standart support
* We require 22.04 to have Qt6 for kiwix-desktop 2.5.0
* There is no fondamental reason to keep supporting 20.04

kelson42:
@rgaudin Wanted to start to move CI to Ubuntu 22.04, starting with libzim. I was kind of optimistic that compiling libzim on Ubuntu 22.04 with all the deps on 20.04 might work. Actually is does not fully work, see [my PR and related CI run](https://github.com/openzim/libzim/pull/977).

Therefore, it is quite clear with need to move deps and base_deps to be compile on Ubuntu 22.04 as well. We could implement #823 straight away, but this would be a kind of a big bang and probably many CI would be broken. This is not really something I'm willing to do. We should be able to have GNU/Linux (base_)deps at the same time for Ubuntu 20.04 and Ubuntu 22.04.

kelson42:
#823 has done most of the job, but we still fail to compile `x86-64_musl` jobs on Ubuntu (Jammy) 22.04. This is still something left and this is not trivial, therefore assigning issue to @veloman-yunkan.

PR #840, opened 2025-06-23
Use zlib-1.3.1 in place of zlib-1.2.12

### LINKED ISSUE #839: Can't compile (zlib) with macOS15.5 (clang17)
On my machine (macOS 15.5, Xcode 16.4), I can't compile `zlib` as the compiler (clang 17) now complains about a 

```
_stdio.h:314:7: error: expected identifier or '(' 
```

It can be fixed by passing `-fno-define-target-os-macros` but ideally this should be conditioned to Clang 17+ ; It might be a problem on earlier clang…

kelson42:
Attempt to fix #839

### DIFF
--- a/kiwixbuild/dependencies/zlib.py
+++ b/kiwixbuild/dependencies/zlib.py
@@ -24,4 +24,8 @@ class zlib(Dependency):
         archives = [src_archive, meson_patch]
         #patches = ['zlib_std_libname.patch']
 
-    Builder = MesonBuilder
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan Imstrongly suspect that his PR is now blocked by https://github.com/kiwix/kiwix-build/issues/823. Would be great if you could fix this issue so we can move to more modern compilation env.

PR #844, opened 2025-09-12
x86-64_musl is now built on jammy

### LINKED ISSUE #823: Remove Ubuntu 20.04 Focal from CI/CD

veloman-yunkan:
Should allow to close #823

### DIFF
--- a/.github/scripts/build_definition.py
+++ b/.github/scripts/build_definition.py
@@ -59,8 +59,8 @@ BUILD_DEF = """
     | jammy     | aarch64_musl_static|        |          | BP        | BP          |               | linux-aarch64-musl      |                        |
     | jammy     | aarch64_musl_mixed | BP     |          |           |             |               | linux-aarch64-musl      |                        |
     | jammy     | aarch64_musl_dyn   | d      |          | B         | B           |               |                         | linux-aarch64-musl-dyn |
-    | focal     | x86-64_musl_static |        |          | BP        | BP          |               | linux-x86_64-musl       |                        |
-    | focal     | x86-64_musl_mixed  | BP     |          |           |             |               | linux-x86_64-musl       |                        |
+    | jammy     | x86-64_musl_static |        |          | BP        | BP          |               | linux-x86_64-musl       |                        |
+    | jammy     | x86-64_musl_mixed  | BP     |          |           |             |               | linux-x86_64-musl       |                        |
     | jammy     | i586_static        |        |          | BP        | BP          |               | linux-i586              |                        |
     | jammy     | i586_dyn           |        |          | B         | B           |               |                         |                        |
     | jammy     | wasm               | dBP    |          |           |             |               | wasm-emscripten         | wasm                   |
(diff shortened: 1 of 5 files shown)


veloman-yunkan:
This PR is ready to have its WIP status removed. The following have to be noted:

1. The main goal was to get rid of the usage of focal in our CI/CD. Although alpine is a more natural build platform for musl, I pursued the seemingly easier solution of migrating the existing `linux-x86_64-musl` build workflows from focal to jammy. If we intend to build musl on alpine, then let's open a ticket for that.
2. To just upgrade the builder for the `linux-x86_64-musl` platform from focal to jammy we need only the first two commits.
3. I also made an attempt (see the 3rd commit) to upgrade the musl toolchain from gcc 10 to gcc 11 (same as the version of gcc on Ubuntu 22.04 jammy). Though it had worked locally on my laptop, the CI jobs ([x86-64_musl_static](https://github.com/kiwix/kiwix-build/actions/runs/17699466053/job/50303215579#logs), [x86-64_musl_mixed](https://github.com/kiwix/kiwix-build/actions/runs/17699466053/job/50303215588#logs)) failed with the following error:

```
[SETUP TOOLCHAINS]
[PREPARE]
prepare sources x86-64_musl :
Stopping build due to errors
  download x86-64_musl : Cannot download URL https://more.musl.cc/11/x86_64-linux-musl/x86_64-linux-musl-cross.tgz
[Errno 101] Network is unreachable
ERROR
```
Are our CI/CD runners configured with whitelisted domains that they are allowed to reach out to?
I believe that this should be easy to fix by hosting the  https://more.musl.cc/11/x86_64-linux-musl/x86_64-linux-musl-cross.tgz on https://dev.kiwix.org like it was done for the previous version.

We need to decide how we want to proceed with this PR.

1. Should we upgrade the musl toolchain?
2. Should we follow up with moving to an alpine builder?

kelson42:
> * Should we upgrade the musl toolchain?

I would preffer, because current one is pretty old. I alos don't reallyblike to rely on "random" binaries.

> * Should we follow up with moving to an alpine builder?

I guess that way woukd be simpler and cleaner and more robust. So good for me.

veloman-yunkan:
> > * Should we follow up with moving to an alpine builder?
> 
> I guess that way would be simpler and cleaner and more robust. So good for me.

I filed #847 for that task and started working on it. However I may propose to complete and merge this PR if it turns out that migrating to an alpine builder requires noticeably more effort.

veloman-yunkan:
@kelson42 The CI run after pushing the last commit is not a correct one - `x86-64_musl` jobs have reused the base_deps archives created by the previous version of the toolchain during the last successful run of CI for this PR. The files `base_deps_jammy_x86-64_musl_static_15.tar.gz` and `base_deps_jammy_x86-64_musl_mixed_15.tar.gz` must be deleted from https://tmp.kiwix.org/ci/ and the CI must be re-run.

Issue #852, opened 2025-09-26
macOS apple_all_static fails when building off base_deps


rgaudin:
Following merge of #850 

I'm on it

rgaudin:
Watch out your PR messages 😆

PR #853, opened 2025-09-28
Bump-up Flatpak KDE dependency to 6.9

kelson42:
Attempt to fix https://github.com/kiwix/kiwix-desktop/issues/1404

Not updating the `base_deps_meta_version` number as this should be done anyway soon, once @rgaudin will have fixed https://github.com/kiwix/kiwix-build/issues/852


veloman-yunkan (commented):
I am not sure what should be reviewed about this PR. I cannot spot any problem but that doesn't mean that this change will go smoothly.

kelson42:
Lets feel lucky then!

Issue #42, opened 2017-05-02
Use the provided CMake file to build pugixml.


mgautierfr:
Pugixml has a Cmake build system who generate a pkg-config file.
We do not need a custom meson file.

See https://github.com/kiwix/kiwix-lib/issues/49

kelson42:
@mgautierfr Is that still a topic?

kelson42:
@mgautierfr We use version 1.2 of pugixml which has been released 5 years ago. We shoukd use latest wich brings as well memory consumption and speed improvements.

eli-schwartz:
pugixml 1.13 is also available in the Meson WrapDB. :) You can just add that instead, and stop maintaining your *own* patch for meson.build support.

kelson42:
@mgautierfr Shoukd we use Meson WrapDB or native tarball with CMake (my preference).

PR #855, opened 2025-09-28
Update dependency to Pugixml 1.15

### LINKED ISSUE #854: Upgrade pugixml
We use a very old version released in 2014... more than 10 years ago. We are more than 10 release behind latest!

kelson42:
Attempt to fix #854

Not updating the base_deps_meta_version number as this should be done anyway soon, once @rgaudin will have fixed https://github.com/kiwix/kiwix-build/issues/852


kelson42:
@veloman-yunkan I have tried to implement #42 at the same time as it woukd simplify many things.... but fails... honestly I doubt the CMakeBuilder is ready (but obviuosly I might be wrong).

kelson42:
@veloman-yunkan Updated the PR, but now it fails mysteriously (at least to me) only for flatpak. Please have a look because now I'm out of my depth. I have the same issue locally, but not really in measure to understand and follow why flatpak works differently from the other platforms!

veloman-yunkan:
> @veloman-yunkan Updated the PR, but now it fails mysteriously (at least to me) only for flatpak. Please have a look because now I'm out of my depth. I have the same issue locally, but not really in measure to understand and follow why flatpak works differently from the other platforms!

@kelson I cannot spot anything in this PR that explains the failure. Will try to debug the flatpak build after I am done with kiwix/libkiwix#1230

kelson42:
@veloman-yunkan I wonder if we are not impacted by a broader issue on system using the new `pkgconf` in place of `pkg-config`. I tested compilation of libzim/zim-tools with g++-15 on brand new Ubuntu 25.10 and, after libzim has been successfuly compiled and installed, zim-tools Meson were not able to identify where is the `.pc` file of libzim.

PR #868, opened 2025-10-24
GitHub Windows runner uses now OpenSSL 3.6.0

kelson42:
Should fix CI/CD errors with Kiwix Desktop Windows packaging


veloman-yunkan (approved):
I was under the impression that OpenSSL 3 is not backward compatible with OpenSSL 1.1.1 but it turns out that it's not so: https://docs.openssl.org/3.0/man7/migration_guide/. In any case, such a change asks for being tested too. Since the build failure is fixed, I don't see why we shouldn't merge but the produced packages still needs to be tested.

PR #872, opened 2025-10-25
Robuster libzim binary archiver

### LINKED ISSUE #871: Nightly `libzim_linux-x86_64-2025-10-25.tar.gz` missing `libzim.so.9.4.0`
From https://download.openzim.org/nightly/2025-10-25/libzim_linux-x86_64-2025-10-25.tar.gz

See https://github.com/kiwix/java-libkiwix/issues/119#issuecomment-3446308340 for more details

I suspect some kind of weakness in kiwix-build because with have version `9.4.0` in libzim and version `9.3.0` declared in `kiwixbuild/versions.py`.

A blocker for libzim release obviously

kelson42:
Fixes #871 by making the source archive packager independant of `main_project_versions`.

### DIFF
--- a/.github/scripts/common.py
+++ b/.github/scripts/common.py
@@ -112,29 +112,15 @@ EXPORT_FILES = {
             ## Linux
             # We need to package all dependencies (`*.a`) on wasm
             *lib_prefix("libzim.a" if COMPILE_CONFIG != "wasm" else "*.a"),
-            *lib_prefix("libzim.so"),
-            *lib_prefix(
-                "libzim.so.{version}".format(version=main_project_versions["libzim"])
-            ),
-            *lib_prefix(
-                "libzim.so.{version}".format(
-                    version=major_version(main_project_versions["libzim"])
-                )
-            ),
+            *lib_prefix("libzim.so*"),
             ## MacOS
-            "lib/libzim.{}.dylib".format(
-                major_version(main_project_versions["libzim"])
-            ),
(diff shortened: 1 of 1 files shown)


kelson42:
> With the proposed change, a usable package with a wrongly labeled binary can be produced, potentially leading to subtle mysteries while testing various packages.

What do you mean with "wrongly labeled binary"? At this stage, the .so filenames are exactly copied like they are. The version taken in anyway the version chosen by kiwix-build, so I don't understand what can be "wrong".

rgaudin:
I understand the goal of the PR but I think we should fix the double-definition of versions instead. We already have a ticket for that.

I'm afraid this could open the door to side effects: are we sure we have symlinks working for all those matching files on all platforms? We may include the binary twice (or more). Creating archives outside GH (fresh state), like on dev machine might mean including several versions. Sure those are not distributed but still, not very clean.
I'm also concerned that we need anyway to have proper versions set for pkgconfig files that are included aside so we do need versions being set correctly.

All in all, this feels like we're saying we're not rigorous enough. I'd prefer we make the task fail if expected file is not present instead.

kelson42:
> I think the PR title is plain wrong as this is about binary archives and not source ones.

fixed

kelson42:
Closing because @veloman-yunkan  and @rgaudin prefer to be stricter here instead of been looser. See https://github.com/kiwix/kiwix-build/issues/871#issuecomment-3485029221 for more details.

PR #876, opened 2025-10-28
Add version number as  property of  Dependency class

### LINKED ISSUE #866: Version number of external dependencies should be specify at the right place
For the moment, we have: 

```
base_deps_versions = {
    "zlib": "1.3.1",
    "lzma": "5.2.6",
    "zstd": "1.5.2",
    "docoptcpp": "0.6.2",
    "uuid": "1.43.4",
    "xapian-core": "1.4.26",
    "mustache": "4.1",
    "pugixml": "1.15",
    "libmicrohttpd": "0.9.76",
    "gumbo": "0.13.1",
    "icu4c": "73.2",
    "libaria2": "1.37.0",
    "libmagic": "5.44",
    "android-ndk": "r21e",
    "org.kde": "6.9",
    "io.qt.qtwebengine": "6.9",
    "zim-testing-suite": "0.8.0",
    "emsdk": "3.1.41",
}
```

in `kiwixbuild/versions.py`

This number should be put as a property of the class `Dependency`.

Roshnicoderr:
Fixes #866

Converted the "base_deps_versions" dictionary into "Dependency" class .

### DIFF
--- a/kiwixbuild/versions.py
+++ b/kiwixbuild/versions.py
@@ -35,23 +35,31 @@ release_versions = {
 # Change this when you change base_deps_versions.
 base_deps_meta_version = "17"
 
-base_deps_versions = {
-    "zlib": "1.3.1",
-    "lzma": "5.2.6",
-    "zstd": "1.5.2",
-    "docoptcpp": "0.6.2",
-    "uuid": "1.43.4",
-    "xapian-core": "1.4.26",
-    "mustache": "4.1",
-    "pugixml": "1.15",
-    "libmicrohttpd": "0.9.76",
-    "gumbo": "0.13.1",
-    "icu4c": "73.2",
-    "libaria2": "1.37.0",
-    "libmagic": "5.44",
-    "android-ndk": "r23c",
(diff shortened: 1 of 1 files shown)


kelson42:
> To me, no clue what the value and why this PR. Like usual it would be better to have first an issue before dropping a PR. But maybe this is just because I'm not a Python dev.

This is a fix for #866, I discovered this afeterward.

kelson42 (changes requested):
@Roshnicoderr Thank you for your PR. Unfortunately this is not what is expected. The version setting code should vaninshed from `versions.py` and the version should be staticaly set in each dependency class/file.

For example around this line for `aria2c` dependency https://github.com/kiwix/kiwix-build/blob/main/kiwixbuild/dependencies/aria2.py#L16.

It is important that this is made in a way that we don't forget to update this string each time we update the upstream tarball URL.

Roshnicoderr:
Hi Kelson, I tried fixing the issue again and have raised a new PR.
Can you please check and let me know if it’s correct or if I missed something?

Issue #873, opened 2025-10-26
Unable to compile libuuid for kiwix-android


kelson42:
```
$ kiwix-build --config android libkiwix
ERROR: The config android cannot be build on host ubuntu.
Select another config or change your host system.
[INSTALL PACKAGES]
 - automake : SKIP
 - libtool : SKIP
 - cmake : SKIP
 - git : SKIP
 - subversion : SKIP
 - ccache : SKIP
 - pkg-config : SKIP
 - gcc : SKIP
 - autopoint : SKIP
 - libbz2-dev : SKIP
 - libmagic-dev : SKIP
SKIP, No package to install.
[SETUP TOOLCHAINS]
[PREPARE]
prepare sources android-ndk :
  download android-ndk : SKIP
  extract android-ndk : SKIP
  patch android-ndk : SKIP
prepare sources pugixml :
  download pugixml : SKIP
  extract pugixml : SKIP
  patch pugixml : SKIP
prepare sources lzma :
  download lzma : SKIP
  extract lzma : SKIP
  patch lzma : SKIP
prepare sources zstd :
  download zstd : SKIP
  extract zstd : SKIP
prepare sources zlib :
  download zlib : SKIP
  extract zlib : SKIP
prepare sources uuid :
  download uuid : SKIP
  extract uuid : SKIP
prepare sources xapian-core :
  download xapian-core : SKIP
  extract xapian-core : SKIP
prepare sources icu4c :
  download icu4c : SKIP
  extract icu4c : SKIP
  patch icu4c : SKIP
prepare sources zim-testing-suite :
  download zim-testing-suite : SKIP
  extract zim-testing-suite : SKIP
prepare sources libzim :
  gitupdate libzim : OK (0.4s)
prepare sources libcurl :
  download libcurl : SKIP
  extract libcurl : SKIP
prepare sources libmicrohttpd :
  download libmicrohttpd : SKIP
  extract libmicrohttpd : SKIP
  patch libmicrohttpd : SKIP
prepare sources mustache :
  download mustache : SKIP
  extract mustache : SKIP
prepare sources libkiwix :
  gitupdate libkiwix : OK (0.4s)
[BUILD]
build android-ndk (android_arm):
  build_toolchain android-ndk : SKIP
  fix_permission_right android-ndk : SKIP
build android-ndk (android_arm64):
  build_toolchain android-ndk : SKIP
  fix_permission_right android-ndk : SKIP
build android-ndk (android_x86):
[...]

PR #882, opened 2025-11-04
Update uuid from 1.43.4 to 1.47.3

vighnesh-sawant:
Addresses #873 
Hey I compiled and checked this does not seem to break anything on android.
Also upstream stopped packaging the libs library separately and combined it into one tar.
It added the --enable-subset flag to include the libs library
https://e2fsprogs.sourceforge.net/e2fsprogs-release.html#1.44.4

> Instead of building the subset e2fsprogs-libs tar file, add a new configure option, --enable-subset. This along with other changes (such as dropping obsolete files that aren't shipped as part of e2fsprogs-X.YY.tar.gz) allows us to be able to build the tarball using the "git archive" command.

### DIFF
--- a/kiwixbuild/dependencies/uuid.py
+++ b/kiwixbuild/dependencies/uuid.py
@@ -8,11 +8,11 @@ class UUID(Dependency):
 
     class Source(ReleaseDownload):
         archive = Remotefile(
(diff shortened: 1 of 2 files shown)


kelson42:
@veloman-yunkan If OK for you I woukd the upload a copy on our drive,

PR #884, opened 2025-11-06
Don't cross-compile manylinux aarch64 libzim

### LINKED ISSUE #881: aarch64 manylinux 9.4.0 is not built / working correctly
See https://github.com/openzim/python-libzim/pull/234#issuecomment-3485607563

This is blocking python-libzim upgrade

rgaudin:
Part of the fix was modifying images on container-images.

Here, it's pretty simple: run the CI/CD for manylinux-aarch64 on an arm64 runner on github and ask kiwix-build to do a `native_mixed` config.

For some reason, compiling UUID dependency (which hasn't changed) wasn't working anymore neither on amd64 nor amd64 due to an `implicit-function-declaration` warning. Both are using newer images so it's likely the compiler is not the same version but that's a loose explanation. I made this warning not fail UUID…

Fixes #881, validated [here](https://github.com/openzim/python-libzim/actions/runs/19134386833/job/54682835032)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
       env:
         COMPILE_CONFIG: ${{matrix.config}}
 
-  ManyLinuxAarch64:
+  LinuxArm64:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -165,6 +163,69 @@ jobs:
       env:
         COMPILE_CONFIG: ${{matrix.config}}
 
+  ManyLinuxAarch64:
```
veloman-yunkan: Same here

veloman-yunkan (approved):

PR #877, opened 2025-10-29
Fix static versions

Roshnicoderr:
Fixes #866

Superseeding #876

PR #891, opened 2025-12-12
Move dependency versions into Dependency classes (#877)

Arpan-Kaur2006:
Fixes #877

Moved all dependency versions from kiwixbuild/versions.py
into their respective Dependency classes as a `version` property,
as suggested in issue #866. This makes version management
more modular and easier to maintain.


kelson42:
@Arpan-Kaur2006 What is done seems correct, but:
* Old place where this is declared is still there
* Code using this version number has not been changed

PR #908, opened 2026-01-14
Add docstring to escape_path utility function

varma1221:
This PR adds a short docstring to the escape_path helper function to clarify
its purpose and behavior when handling filesystem paths with spaces.

No functional or behavioral changes are introduced.

### DIFF
--- a/kiwixbuild/utils.py
+++ b/kiwixbuild/utils.py
@@ -47,6 +47,11 @@ regex_space = re.compile(r"((?<!\\) )")
 
 
 def escape_path(path):
+    """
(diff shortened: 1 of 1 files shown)


veloman-yunkan (commented):
I am not sure that the proposed docstring helps. Anyone with minimal experience in python could easily deduce the same information from the current implementation of the function. On the other hand, the function name suggests that it should do more work than escaping the spaces alone, thus the current implementation may be deficient and even buggy ~(e.g. on input like `r'abc\ def'` that looks like an already backslash escaped string)~.

PR #917, opened 2026-02-16
import buildenv explicitly in __init__.py to prevent dependency errors

lumpy72006:
`kiwixbuild/__init__.py` uses `buildenv` but does not import it.

Currently, it works due to a side-effect where importing the `dependencies` package  loads `buildenv` and binds it to the kiwixbuild scope. If dependencies is refactored or import order changes, `__init__.py` crashes with a `NameError` and it's been confusing trying to figure out how this worked in the first place


rgaudin (approved):
Indeed, thank you.

Issue #845, opened 2025-09-13
Upgrade obselete AppImageKit


kelson42:
See https://github.com/kiwix/kiwix-build/pull/843 for more details

PR #929, opened 2026-05-08
Fix CI/CD after bringing libtorrent to kiwix-desktop

### LINKED ISSUE #1488: Introduction of libtorrent has broken default compilations
See https://github.com/kiwix/kiwix-build/actions/runs/25531714590

It seems that #1469 has broken compilations on kiwix-build

The optional compilation argument activating libtorrent support seems to not apply properly everywhere

veloman-yunkan:
Should fix kiwix/kiwix-desktop#1488

The biggest pitfall with this PR is that the builder for `kiwix-desktop` is upgraded from `jammy` to `noble`, the most undesirable consequence of it being that the `kiwix-desktop` appimages produced by our CI/CD will no longer run on `jammy`.

### DIFF
--- a/.github/scripts/build_definition.py
+++ b/.github/scripts/build_definition.py
@@ -39,7 +39,8 @@ BUILD_DEF = """
     | jammy     | flatpak            |        |          |           |             | BP            |                         |                        |
     | jammy     | native_static      | d      | d        | dBPSD     | dBPSD       |               | linux-x86_64            | linux-x86_64-static    |
     | jammy     | native_mixed       | BPS    | BPS      |           |             |               | linux-x86_64            |                        |
-    | jammy     | native_dyn         | d      | d        | dB        | dB          | dBPS          |                         | linux-x86_64-dyn       |
+    | jammy     | native_dyn         | d      | d        | dB        | dB          |               |                         | linux-x86_64-dyn       |
+    | noble     | native_dyn         |        |          |           |             | dBPS          |                         | linux-x86_64-dyn       |
     # libzim CI is building alpine_dyn but not us
     | jammy     | android_arm        | dBP    | dBP      |           |             |               | android-arm             | android-arm            |
     | jammy     | android_arm64      | dBP    | dBP      |           |             |               | android-arm64           | android-arm64          |
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -137,7 +137,7 @@ jobs:
       KIWIX_FILE_UPLOAD_SSH_KEY_PATH: /tmp/kiwix_file_upload_key
     runs-on: ubuntu-22.04
     container:
-      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:2025-11-05"
+      image: "ghcr.io/kiwix/kiwix-build_ci_${{matrix.image_variant}}:2026-04-25"
       options: "--device /dev/fuse --privileged"
     steps:
     - name: Checkout code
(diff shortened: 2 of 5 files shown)


kelson42:
We don't need to have bittorrent support in Flatpak now. Having Bittorrent support in kiwix-desktop is anyway temporary. We should focus on proving the feature works and the move everything to libkiwix. Goalis to reach that point as quickly as possible.

veloman-yunkan:
I tried one last time. If it fails too, I will instead disable `libtorrent` in the flatpak build. In any case it was a useful first exposure to flatpak builds, as well some knowledge gained towards making `libtorrent` a dependency of `kiwix-build` later.

veloman-yunkan:
> I tried one last time. If it fails too, I will instead disable `libtorrent` in the flatpak build. In any case it was a useful first exposure to flatpak builds, as well some knowledge gained towards making `libtorrent` a dependency of `kiwix-build` later.

That "last" attempt failed, but in a way that was easier to fix than to have `libtorrent` disabled in a flatpack build (which required a change in another (`kiwix-desktop`) repository).

kelson42:
> > I tried one last time. If it fails too, I will instead disable `libtorrent` in the flatpak build. In any case it was a useful first exposure to flatpak builds, as well some knowledge gained towards making `libtorrent` a dependency of `kiwix-build` later.
> 
> That "last" attempt failed, but in a way that was easier to fix than to have `libtorrent` disabled in a flatpack build (which required a change in another (`kiwix-desktop`) repository).

For the last appimage problem, we are impacted by https://github.com/kiwix/kiwix-build/issues/845 ?

veloman-yunkan:
> For the last appimage problem, we are impacted by #845 ?

Likely. Upgrading `linuxdeploy` and `linuxdeploy-plugin` to most recent versions, and switching from `obsolete-appimagetool` to the most recent (non-continuous) release of `appimagetool` made the `create_kiwix-desktop_appImage.sh` script run to the end, however it didn't produce a usable appimage of `kiwix-desktop`. That is exactly the kind of problem [I was reluctant to face](https://github.com/kiwix/kiwix-build/pull/843#issuecomment-3288245208) before running into a pressing need. Now we have to deal with it.

veloman-yunkan:
Fortunately, it proved to be a Qt5-vs-Qt6 issue. Forcing Qt6 (on my machine) produced an appimage from which kiwix-desktop could be run. So I would be surprised if CI fails or doesn't deliver a usable appimage.

kelson42:
> Fortunately, it proved to be a Qt5-vs-Qt6 issue. Forcing Qt6 (on my machine) produced an appimage from which kiwix-desktop could be run. So I would be surprised if CI fails or doesn't deliver a usable appimage.

If I get it right, I have to cross all my fingers that next CI run works?

veloman-yunkan:
appimage created on Ubuntu `noble` doesn't run on Ubuntu `jammy` (because of `glibc` version mismatch). That was not much of a surprise, but I had forgotten that that was the main reason for sticking to as old platform for our official builds as reasonably possible.

veloman-yunkan:
> > Fortunately, it proved to be a Qt5-vs-Qt6 issue. Forcing Qt6 (on my machine) produced an appimage from which kiwix-desktop could be run. So I would be surprised if CI fails or doesn't deliver a usable appimage.
> 
> If I get it right, I have to cross all my fingers that next CI run works?

Do you have a black belt at finger crossing? The CI is green!

But please don't merge this PR before carefully considering the negative ramifications of the approach taken to address the broken CI.

kelson42:
> But please don't merge this PR before carefully considering the negative ramifications of the approach taken to address the broken CI.

What do you mean exactly? What is left to do to your opinion?

veloman-yunkan:
> > But please don't merge this PR before carefully considering the negative ramifications of the approach taken to address the broken CI.
> 
> What do you mean exactly? What is left to do to your opinion?

If we are going to [drop support for jammy](https://github.com/kiwix/kiwix-build/pull/929#issuecomment-4415610578) we must do it consciously rather than because something was overlooked.

Issue #946, opened 2026-06-14
Build binaries --without-writer


kelson42:
We can now build libzim without the whole writer code. See https://github.com/openzim/libzim/pull/1093

We ahould use it for:
* Android
* wasm
* if possible for Apple libkiwix (used in kiwix-apple repo)

kelson42:
I guess, here only Apple is missing?

PR #947, opened 2026-06-15
Building libzim without writer on android and wasm

veloman-yunkan:
Fixes part of #946

Doing the same for iOS is somewhat problematic because of #948


veloman-yunkan:
I propose pursuing the same change for iOS in a separate PR, after #948 is fixed

PR #950, opened 2026-06-17
Clean-up of the ios vs macos mess

### LINKED ISSUE #948: Mess with ios vs macos
It looks like support for **macos** was added to `kiwix-build` as an enhancement over existing support for **ios**. As a result, it seems like distinction between ios and macos is not properly made.

For example, `AppleConfigInfo` defines `build` as `"ios"`

https://github.com/kiwix/kiwix-build/blob/06190d07074a6284626dd8087c66d1a3578b69d3/kiwixbuild/configs/ios.py#L12-L13

which is inherited by not only by `iOSArm64`, `iOSx64Simulator`, and `iOSArm64Simulator`

https://github.com/kiwix/kiwix-build/blob/06190d07074a6284626dd8087c66d1a3578b69d3/kiwixbuild/configs/ios.py#L155-L179

but also by `macOSArm64`, `macOSArm64Mixed`, `macOSAmd64` and ``macOSAmd64Mixed`

https://github.com/kiwix/kiwix-build/blob/06190d07074a6284626dd8087c66d1a3578b69d3/kiwixbuild/configs/ios.py#L182-L219

This complicates differentiating real ios from macos in kiwix-build logic.

veloman-yunkan:
Fixes #948


rgaudin (approved):
LGTM. Haven't tested but I see all tests passed and I've compared the output xcframework from this branch with the previous one. That looks OK 👍

PR #951, opened 2026-06-18
Building libzim without writer on iOS

### LINKED ISSUE #946: Build binaries --without-writer

veloman-yunkan:
Fixes #946

### DIFF
--- a/kiwixbuild/dependencies/libzim.py
+++ b/kiwixbuild/dependencies/libzim.py
@@ -32,7 +32,7 @@ class Libzim(Dependency):
         @property
         def configure_options(self):
             configInfo = self.buildEnv.configInfo
-            if configInfo.build in ("android", "wasm"):
+            if configInfo.build in ("android", "ios", "wasm"):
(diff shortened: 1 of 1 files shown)


veloman-yunkan:
Standalone `ios_*` and `macos_*` configs seem to work fine, however `apple_all_static` config (through which the release artifacts are created ) is problematic. When building that configuration need to

1. disable writer for `ios_*` subconfigs and
2. when creating the xcframework tree, copy header files from the respective build output (currently it is assumed that the header file set is the same across all subconfigs, and so they are taken from one of the macos configs).

kelson42:
@veloman-yunkan Any news here? What is the status?

soloturn:
> @veloman-yunkan Any news here? What is the status?

made a proposal on top of here @kelson42 @veloman-yunkan

PR #959, opened 2026-08-05
Include break iteration data in our ICU build

veloman-yunkan:
Needed for openzim/zim-tools#499

Summary of changes:

- Updated the ICU data filter so that break iteration data is also pulled in in our non-Windows builds
- In Windows build `pkgconfig` file is generated for the `icu-uc` package too
- Some clean-up


veloman-yunkan:
This PR has been validated in openzim/zim-tools#499 (see https://github.com/openzim/zim-tools/pull/499/changes/533eac92138bbc67e186564e6fb6af26a2861b73 and [its CI run](https://github.com/openzim/zim-tools/actions/runs/31087559120)).

PR #967, opened 2026-09-11
Release libzim 9.8.2-1

veloman-yunkan:
This build of libzim 9.8.2 re-enables optimized compilation of Xapian (see #966).

### DIFF
--- a/kiwixbuild/versions.py
+++ b/kiwixbuild/versions.py
@@ -8,26 +8,26 @@ main_project_versions = {
     "kiwix-desktop": "2.5.1",
 }
 
-# This dictionnary specify what we need to build at each release process.
+# This dictionary specifies what we need to build at each release process.
 # - Values are integer or None
 # - If a project is not in the dict (or None), the project is not released.
 # - If release_versions[project] == 0, this is the first time the project is
-#   build for this release, so publish src and build archives.
+#   built for this release, so publish src and build archives.
 # - If release_versions[project] > 0, release only the build archive with a
 #   build postfix.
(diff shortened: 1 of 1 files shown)


kiwixbuild/versions.py
```diff
@@ -8,26 +8,26 @@
...
 #     `(was ...)`.
 #    - Else, increment the value. If no value was present, see `(was ...)`.
 
 release_versions = {
-    "libzim": None,  # Depends of base deps (was 0)
-    "libkiwix": None,  # Depends of libzim (was 2)
-    "kiwix-tools": None,  # Depends of libkiwix and libzim (was 0)
-    "zim-tools": 0,  # Depends of libzim (was None)
-    "kiwix-desktop": None,  # Depends of libkiwix and libzim (was 0)
+    "libzim": 1,  # Depends on base deps (was None)
```
kelson42, line 26: Here it should be `was 0`

veloman-yunkan: It almost doesn't matter. I mimicked how it was done in f5b1ff5ef50266086c8a328706eba56c3ffddadb

kelson42: yes, I made it falsy as well... it does potentially matter at the next iteration.

veloman-yunkan: > it does potentially matter at the next iteration.

No, it doesn't. The previous value matters only if the current value is `None` (that's pretty much the only situation when it makes sense to preserve the previous value lest one should be forced to study the git history).

